PR #8, opened 2020-04-10
fix issue with ip4 addresses below 128.0.0.0

steff333:
If IP addresses are below 128.0.0.0 the corresponding integer value is not below 0, so adding 2 **32  to the value results in illegal values and the reading of values crashes

### DIFF
--- a/luxtronik/datatypes.py
+++ b/luxtronik/datatypes.py
@@ -82,11 +82,15 @@ class IPAddress(Base):
     measurement_type = "ipaddress"
 
     def _to(self, v):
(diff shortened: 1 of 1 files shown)


kbabioch:
Okay, this might explain the change needed to the `_from` function, i.e. for values received from the heat pump. For the `_to` function I still don't get it, since you're adding `2**32` to the value before passing it into the `ipaddress` library. This will result in an Error.

If, at all needed, the value needs to be added to the resulting calculation of `ipaddress`.

I will try to reproduce this later on. Do you happen to know where this report was coming from and/or which parameter was affected by this?

PR #14, opened 2020-06-06
Remove logging config as it breaks projects one

tafli:
Having `logging.basicConfig()` in a library will break ones own config if the library is imported before owns config.

### DIFF
--- a/luxtronik/__init__.py
+++ b/luxtronik/__init__.py
@@ -7,7 +7,6 @@ from luxtronik.calculations import Calculations
 from luxtronik.parameters import Parameters
 from luxtronik.visibilities import Visibilities
 
-logging.basicConfig(level="WARNING")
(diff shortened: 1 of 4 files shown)


Bouni:
Thanks for your PR. I'm not very experienced with logging and never had problems before :man_shrugging:

tafli:
Thanks, no problem at all!

I use to have my own logging config at the start:

`logging.basicConfig(
    level=LOGLEVEL,
    format="%(asctime)s.%(msecs)03d %(name)s %(levelname)-8s %(message)s",
    datefmt="%Y-%m-%d %H:%M:%S")`

So in this case, the import `from luxtronik import Luxtronik` has to come after the `basicConfig` as it would already be set by your lib.

I'll also write a PR with some log levels on INFO instead on DEBUG, imo (like: _Connected to Luxtronik heatpump_). This way one can set output if needed.

PR #15, opened 2020-06-13
Add unknown visibilities and parameters

WhistleMaster:
Add unknown:
- visibilities: 346, 347 and 348 
- parameters: 1120, 1121 and 1122


Bouni:
Thanks for your PR!

PR #19, opened 2020-11-20
Update parameters.py

WhistleMaster:


WhistleMaster:
Add unknown visibilities, parameters and calculation for V3.85.6 fw

Bouni:
Thanks for the PR!

Issue #36, opened 2023-01-11
Optimizing data request


BenPru:
I have some challenges to make my ha luxtronik integration stable:

- Not all write commands arrive at the heatpump. [#54](https://github.com/BenPru/luxtronik/issues/54)
- Written values have a delay. [#48](https://github.com/BenPru/luxtronik/issues/48)
- V2.88.1 firmware crashes when requesting or writing many times in a short period of time.

I have figured out that the firmware has a problem with the reconnects on the socket.
So I have made some changes in the core module:

- Added a thread lock to only run one command (read/write) at a time.
- Close socket connection only in the desctructor.
- Read the values after write command to get the changes promptly.

At the moment I'm still testing, but you might want to take a look:
https://github.com/Bouni/python-luxtronik/commit/0daf0e6b7caf64ca126f0839d4661d1a8d15bcf3

kbabioch:
To my understanding @BenPru is working on this currently in: https://github.com/BenPru/python-luxtronik/tree/Optimizing-data-reques

BenPru:
@kbabioch I have fixed the linter messages. But the workflow is using the old source.
https://github.com/Bouni/python-luxtronik/actions/runs/3931922636/jobs/6724000633#step:5:1
How can I force a re run with the current source?

kbabioch:
Is this still an issue or have you fixed it by now? The merge request #55 looks fine as far as I can tell in terms of linting?

Bouni:
@kbabioch Did you test the changes in #55 ? I cannot test it at the moment ...

BenPru:
After the merge it is fixed for me.

PR #37, opened 2023-01-11
Datatype minute, change datatypes

BenPru:
### Changed - Make writeable and change datatype
- ID_Einst_LGST_akt --> Celsius
- ID_Einst_TAbsMin_akt --> Celsius
- ID_Einst_Minimale_Ruecklaufsolltemperatur --> Celsius
- ID_Einst_HRHyst_akt --> Kelvin
- ID_Einst_TRErhmax_akt --> Kelvin
- ID_Einst_Popt_Nachlauf_akt --> Minutes
- ID_Einst_Freigabe_Zeit_ZWE --> Minutes

### Added
* New datatype minutes.

### DIFF
--- a/.github/workflows/pylint.yml
+++ b/.github/workflows/pylint.yml
@@ -7,7 +7,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10"]
+        python-version: ["3.9", "3.10", "3.11"]
     steps:
     - uses: actions/checkout@v3
     - name: Set up Python ${{ matrix.python-version }}


luxtronik/__init__.py
```diff
@@ -1,21 +1,44 @@
...
+import time
 
 from luxtronik.calculations import Calculations
 from luxtronik.parameters import Parameters
 from luxtronik.visibilities import Visibilities
+# endregion Imports
 
 LOGGER = logging.getLogger("Luxtronik")
 
+def is_socket_closed(sock: socket.socket) -> bool:
```
kbabioch, line 19: Looking at the description of this pull request and the included changes, I can see that something is not correct.

The descriptions is talking about new datatypes, while the changes are also changing/modifying/improving the socket handling.

From my point of view this should be separated into multiple pull requests. It seems to me that for the datatypes only changes in luxtronik/datatypes.py and luxtronik/parameters.py are needed.

The changes in luxtronik/__init__.py are most likely referring to #36 and should be handled on their own.

luxtronik/__init__.py
```diff
@@ -58,9 +97,14 @@ def write(self):
             LOGGER.debug("Command %s", cmd)
             val = struct.unpack(">i", self._socket.recv(4))[0]
             LOGGER.debug("Value %s", val)
-        self._disconnect()
-        # flush queue after writing all values
+        # Flush queue after writing all values
         self.parameters.queue = {}
+        # Give the heatpump a short time to handle the value changes/calculations:
+        time.sleep(0.2)
```
kbabioch, line 103: Ideally there should be no magic hardcoded values within the code itself. Rather this should be extracted into a "constant" and be named something like `WAIT_TIME_WRITE_PARAMETER`.

luxtronik/__init__.py
```diff
@@ -24,28 +47,44 @@ def __init__(self, host, port, safe=True):
...
+            self._read()
+
+    def _read(self):
         self._read_parameters()
         self._read_calculations()
         self._read_visibilities()
-        self._disconnect()
 
-    def write(self):
-        """Write patameter to heatpump."""
```
kbabioch, line 47: Seems like you're removing the docstring here. Is this by accident or on purpose?

luxtronik/datatypes.py
```diff
@@ -86,6 +86,12 @@ class Seconds(Base):
     measurement_type = "seconds"
 
 
+class Minutes(Base):
+    """MinutesSeconds datatype, converts from and to Minutes."""
```
kbabioch, line 90: `Seconds` is probably a leftover in this string, should be removed:

```suggestion
    """Minutes datatype, converts from and to Minutes."""
```

luxtronik/__init__.py
BenPru: Yes, sorry. The changes in __init__.py should not included in this pr. I have pushed it after the pr...
Can I revert this commit or delete it?

luxtronik/__init__.py
BenPru: accident - function changed from write to _write.
Write see L63 has docstring

luxtronik/__init__.py
kbabioch: Technically the correct thing to fix this is to rebase it. Basically you have to do something along those lines:

git clone git@github.com:Bouni/python-luxtronik.git
cd python-luxtronik
git remote add benpru git@github.com:BenPru/python-luxtronik.git
git checkout -b benpru/main
git rebase -i origin/main 

Here you could delete the commits that are not relevant.

However the main problem is/was that you've created a MR from the `main` branch of your repository. Typically you should try to avoid that and create merge requests from "feature branches" into `main`. When `main` is the source, you cannot easily add additional commits.

Bouni:
@BenPru Something went clearly wrong here 😬 I guess commit [13ba1d1](https://github.com/Bouni/python-luxtronik/pull/37/commits/13ba1d1dcf1796723d9ab93b8589ec499c1feeef) is what you inteded to send in as a PR.

Can you please fix this?

BenPru:
> Can you please fix this?

Fixed in #41

PR #42, opened 2023-01-15
Add support for access level parameter

kbabioch:
This adds full support for parameter `107`. This parameter is used internally
to define the access level to all kind of additional data that is not available
for typical users.

The access level can be changed locally by entering special codes, which are
well known among most advanced users:

- `9445` -> `installer`
- `7113` -> `after sales service`

There is an additional mode that can be enabled by inserting a specially
crafted USB stick into the USB interface of the controller. This will
grant the user `manufacturer` access level.

In the end, though, these hurdles are all falling back to using parameter `107`
to define the access value.

Therefore the current access level can be read from this parameter and be
translated into a string via the `AccessLevel` datatype.

Additionally the current access level can be set by writing the correct value
to this specific parameter, essentially overcoming the access hurdles.

This has been tested locally by me and it works fine here. The strings are
those that are being used by the firmware when using `English` as language.

In German the values are: `Benutzer`, `Installateur`, `Kundendienst`, `Werk`.

### DIFF
--- a/luxtronik/datatypes.py
+++ b/luxtronik/datatypes.py
@@ -530,6 +530,19 @@ class SecOperationMode(SelectionBase):
     }
 
 
+class AccessLevel(SelectionBase):
+    """AccessLevel datatype, converts from and to list of AccessLevel codes"""
+
+    measurement_type = "selection"
+
+    codes = {
+        0: "user",
+        1: "after sales service",
+        2: "manufacturer",
+        3: "installer",
+    }
+
+
 class Unknown(Base):
     """Unknown datatype, fallback for unknown data."""
 
--- a/luxtronik/parameters.py
+++ b/luxtronik/parameters.py
@@ -4,6 +4,7 @@
 import logging
 
 from luxtronik.datatypes import (
+    AccessLevel,
     Kelvin,
     Celsius,
     CoolingMode,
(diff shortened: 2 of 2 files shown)


gerw:
Is it really enough to set parameter #107? In the function `ZugriffstufeAndern` in the firmware, there are also calls to `SetAccess` and to `SetZugriff`. Both methods set `(&AktuelleZugriffstufenParameter)[param_number]` for certain `param_number`s. However, I am not sure where this is used. I found some access to this in `GetZugriff` but this method seems not to be called at many places in the code.

kbabioch:
Yes, pretty sure it is sufficient to only set this parameter. That's what I did. Tested this quite extensively. Also when entering the correct codes or inserting the special USB stick, this is the only parameter that is being changed. Have a look at the event log in my screenshot.

gerw:
You prepared an usb stick for factory access?

kbabioch:
Yes, although I don't want to adorn myself with borrowed plumes, since the script was originally written by someone else.

Bouni:
@kbabioch looks good to me but I wonder if the access levels are in the correct order?

```
codes = {
        0: "user",
        1: "after sales service",
        2: "manufacturer",
        3: "installer",
    }
```

The following would meake more sense to me:

codes = {
        0: "user",
        1: "after sales service",
        2: "installer",
        3: "manufacturer",
    }
```

But  what I've seen in the Android app would totally give way for such a unintuitive order 😆

gerw:
Yes, I can confirm that this is correct.

In the firmware, either a bitmask is used or the value of `Einst_Zugangscode` is queried explicitly.

kbabioch:
It's indeed this "messy". In terms or privileges it is like this:

`user` < `installer` < `after sales support` < `manufacturer`

The codes are not coherent with that, but that's why we have the "code mapping".

Bouni:
🤷🏽‍♂️ OK, so i guess than I'll merge it 😅

PR #50, opened 2023-01-16
Improve linting of source files

kbabioch:
This consolidates the `pylint.yml` workflow into the `lint.yml` workflow, as both are dealing with source file linting and the latter is already containing jobs for multiple aspects (`yamllint`, `markdownlint`). As such it makes sense to combine all of it into one workflow with multiple jobs.

To ensure that linting is run on incoming pull requests prior to merging the changes into `main`, an additional trigger source is defined (`pull_request`).

Additionally some minor adjustments are made to make all of the actions more consistent with each other.


kbabioch:
As you can see in this pull request, the linting checks are now being run in the pull request already and are represented as `Checks`. With this we can try to enforce some consistency through the repo with any newly incoming pull requests.

Bouni:
Nice, I love actions that improve consistency 😍

kbabioch:
> Nice, I love actions that improve consistency heart_eyes

Well, let's see how long it takes until the first one gets annoyed by too strict rules :-).

PR #55, opened 2023-01-16
Optimize data requests

### LINKED ISSUE #36: Optimizing data request

BenPru:
Resolves #36

### DIFF
--- a/luxtronik/__init__.py
+++ b/luxtronik/__init__.py
@@ -33,7 +33,7 @@ def is_socket_closed(sock: socket.socket) -> bool:
         return False  # socket is open and reading from it would block
     except ConnectionResetError: # pylint: disable=W0703
         return True  # socket was closed for some other reason
-    except Exception as err:
+    except Exception as err: # pylint: disable=W0703
         LOGGER.exception(
             "Unexpected exception when checking if a socket is closed", exc_info=err
         )
--- a/luxtronik/__init__.py
+++ b/luxtronik/__init__.py
@@ -35,7 +35,7 @@ def is_socket_closed(sock: socket.socket) -> bool:
         return True  # socket was closed for some other reason
     except Exception as err: # pylint: disable=W0703
         LOGGER.exception(
-            "Unexpected exception when checking if a socket is closed", exc_info=err
+            "Unexpected exception when checking if socket is closed", exc_info=err
         )
         return False
     return False
--- a/luxtronik/__init__.py
+++ b/luxtronik/__init__.py
@@ -42,7 +42,7 @@ def is_socket_closed(sock: socket.socket) -> bool:
 class Luxtronik:
     """Main luxtronik class."""
 
-    def __init__(self, host, port, safe=True):
+    def __init__(self, host, port=8889, safe=True):
         self._lock = threading.Lock()
         self._host = host
         self._port = port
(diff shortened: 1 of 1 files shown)


luxtronik/__init__.py
```diff
@@ -24,28 +54,59 @@ def __init__(self, host, port, safe=True):
...
     def read(self):
         """Read data from heatpump."""
-        self._connect()
+        self._read_write(write=False)
+
+    def write(self):
+        """Write parameter to heatpump."""
+        self._read_write(write=True)
+
+    def _read_write(self, write=False):
```
kbabioch, line 74: I'm not particularly good with names and the change itself looks good. Just wondering whether this function could/should be renamed to something like `_read_after_write`.

To me `_read_write()` doesn't express what is actually happening, also it should be more like `_write_read()`, since values are first written, then read back :-).

luxtronik/__init__.py
```diff
@@ -58,9 +119,15 @@ def write(self):
...
-        self._disconnect()
-        # flush queue after writing all values
+        # Flush queue after writing all values
         self.parameters.queue = {}
+        # Give the heatpump a short time to handle the value changes/calculations:
+        time.sleep(WAIT_TIME_BETWEEN_WRITE_READ_PARAMETER)
+        # Read the new values based on our parameter changes:
+        self._read_parameters()
+        self._read_calculations()
+        LOGGER.warning("New: '%s'", self.calculations.get(12))
```
kbabioch, line 129: I don't understand why reading and logging calculation 12 (`ID_WEB_Sollwert_TRL_HZ`) in a generic write method is needed.

Could this be a leftover from some testing?

BenPru:
> @BenPru can you confirm this is working as expected?

Yes, its working very good for me. Should I merge?

luxtronik/__init__.py
```diff
@@ -1,21 +1,51 @@
...
+        data = sock.recv(16, socket.MSG_DONTWAIT | socket.MSG_PEEK)
+        if len(data) == 0:
+            return True
+    except BlockingIOError:
+        return False  # socket is open and reading from it would block
+    except ConnectionResetError: # pylint: disable=W0703
+        return True  # socket was closed for some other reason
+    except Exception as err: # pylint: disable=W0703
+        LOGGER.exception(
+            "Unexpected exception when checking if a socket is closed", exc_info=err
```
kbabioch, line 38:
```suggestion
            "Unexpected exception when checking if socket is closed", exc_info=err
```

Bouni:
@BenPru Yes, feel free to merge!

Issue #59, opened 2023-01-18
Add dump functionality


Bouni:
A lot of users ask how to find certain values and I always point to a [gist](https://gist.github.com/Bouni/cf0dc2c7c53c32deeb264803ec350e5c) where I put a script that does this.

Maybe we should integrate this functionality into the module itself so that we can cal it like `python luxtronik dump` or so.

kbabioch:
I would rather suggest to add a `scripts` directory to the repository and put the script there, e.g. something like `scripts/dump-luxtronik.py`. To my understanding this is a script making use of the library and not a classical Python module.

Bouni:
Suggesting `python luxtronik dump` was by intention.

That way a user can simply install the module with `pip install luxtronik` and then call the function.
If we just bundle the script the user cannot simply call it without knowing the install path, right?  

Maybe thats the first step towards some kind of CLI (not sure if that makes sense tbh)

kbabioch:
Having a CLI later is a great idea, i.e. you could interactively change some values and get the look and feel like sitting in front of the controller.

Regarding having this as a Python module: Personally, I don't like that too much. I expect users of a Python library to locate a script and run it. Also its a question of packaging, i.e. you could easily put the script into `PATH`, so you don't have to look for it.

AFAIK most Python libraries provide dedicated scripts, rather than Python modules than are imported and run some magic. Besides that: How would you specify IP address and port when running `python luxtronik dump`?

Bouni:
I just put to gether a really quick and dirty cli-draft: 

The essential changes are here: https://github.com/Bouni/python-luxtronik/blob/cli-draft/setup.py#L17
And here: https://github.com/Bouni/python-luxtronik/blob/cli-draft/luxtronik/__main__.py

With this you pip install the luxtronik module and can run a `luxtronik --ip 192.168.88.11 --port 8889 dump` without the need for `python -m` or similar as prefix.

Disclaimer, I've never really used argparse, maybe there's a more elegant solution for this.

kbabioch:
While #61 addresses the issue, I don't think we had discussed and aligned the route here yet. So I'm re-opening this issue in order to revisit it.

Personally I think scripts along with the package are "good enough", but I can also see some advantages in the proposal from @Bouni, so I'm not against it and also fine with getting rid of my pull request #61 in favor of another solution.

[...]

PR #61, opened 2023-01-18
Add script to dump all values from Luxtronik controller

kbabioch:
This adds a script that will allow users to dump all known values from
the Luxtronik controller, including:

- parameters
- calculations
- visibilities

The script itself was taken from a [Gist][gist].

In addition to that a few changes have been made:

- Changes to code to stay compliant with `pylint`
- Implementation of argument parsing via `argparse`

This change closes https://github.com/Bouni/python-luxtronik/issues/59.

[gist]: https://gist.github.com/Bouni/cf0dc2c7c53c32deeb264803ec350e5c


BenPru (approved):
Great

PR #62, opened 2023-01-18
Make Kelvin datatype consistent to Celsius

### LINKED ISSUE #58: Celsius vs. Kelvin datatypes convert differently
The datatypes `Kelvin` and `Celsius` seem to be implemented quite similar, yet a little bit different:

https://github.com/Bouni/python-luxtronik/blob/d9f0e12bb02400ed9d1d0f8f77f3bb426185f78e/luxtronik/datatypes.py#L53-L62

https://github.com/Bouni/python-luxtronik/blob/d9f0e12bb02400ed9d1d0f8f77f3bb426185f78e/luxtronik/datatypes.py#L134-L143

As you can see in case of `Celsius` the value in `to_heatpump` is first cast into a `float` before being multiplied and then converted back into an `int`.

In case of `Kelvin` there is no explicit cast into `float`.

Not sure if it makes a difference in reality (can't think of such a case right now), but also not sure if this is wanted or just overlooked.

Should this be changed?

kbabioch:
This is a small change to the Kelvin datatype to make it consistent with the Celsius datatype, since they were inconsistent previously.

This will allow us to refactor the code later on more easily (e.g. have some kind of temperature super class, etc.).

This addresses and closes #58.

### DIFF
--- a/luxtronik/datatypes.py
+++ b/luxtronik/datatypes.py
@@ -140,7 +140,7 @@ class Kelvin(Base):
         return value / 10
 
     def to_heatpump(self, value):
-        return int(value * 10)
+        return int(float(value) * 10)
 
 
 class Pressure(Base):


Bouni:
Perfect, thanks for the PR!

PR #64, opened 2023-01-19
Implement unit testing for datatypes module

kbabioch:
This implements some basic unit tests. Having unit tests is considered to be good practice and allows us to refactor the code while reducing the risk for regressions and silly mistakes. The test suites are based on `pytest`, which is the de-facto standard for unit tests in Python.

The pull request consists of multiple changes / commits:

- Implementation of actual unit tests in `tests/test_datatypes.py`. This is a first basic test suite for the `datatype` module. It is not covering every edge case, and is more of a starting point that can be used to elaborate later on. Other modules also need to have unit tests implemented, which will follow in additional pull requests later on.

- Implementation of `pytest` as GitHub action: This will run the test suites automatically whenever new code is pushed or when existing code is pull requested.

For now I had to disable the coverage reporting via comment due to missing permission issues. Potentially that is something that will work when @Bouni will merge it and it is running in the context of this original repository (and not my forked one).

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -40,6 +40,8 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
+      - name: Install pytest
+        run: pip install pytest
       - name: Install pylint
         run: pip install pylint
       - name: Run pylint
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -17,6 +17,14 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
       - name: Install pytest
-        run: pip install pytest
+        run: pip install pytest pytest-cov
       - name: Run pytest
-        run: PYTHONPATH=. pytest
+        run: >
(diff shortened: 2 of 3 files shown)


kbabioch:
Seems like there are permission issues when trying to create a pull request from my personal repository into this one. Need to look into how to resolve this.

Hopefully the change itself should be fine.

kbabioch:
I've rebased on current `master`. For now tests for `datatypes` are implemented with full coverage. I plan to use this as basis for further refactoring.

@Bouni and @BenPru could you have a look and review this? Unit tests for other modules are missing, but we need to start somewhere and implementing unit tests for everything at once is quite some work. This way we will have unit tests for the most essential module at least, and whenever we're going to refactor some code, we will hopefully ensure that we don't run into regressions.

kbabioch:
Also, @Bouni I had to disable the comment action due to missing permission. I assume those will start working, once this is running in the context of this repository, so I will try to re-enable them after the merge ;-).

kbabioch:
Rebased once more on top of current master to make sure changes from #79 are included.

Bouni (approved):

PR #65, opened 2023-01-19
Automatically bump version on release

Bouni:


Bouni:
I let the `cat setup.py` in there on purpose, if something doesn't work as expected we have a good debugging help and it doesn't hurt in my opinion.

kbabioch (approved):
Changes look good to me. Potentially the escaped double-quotes could be avoided by using single quotes, but that's not too important.

Bouni:
The escaped double quotes are by intention, we need the outer double quotes because the `${VERSION}` within the string wouldn't be evaluated otherwise and havin single quotes for the inner part (the escaped ones) would result in singlequotes in the `setup.py` where all the other values have double quotes. And that, my friend, would trigger my inner Monk 😆

kbabioch:
Technically the single quotes could be used as outer quotes (on shell level), but nevermind, it works and is not relevant anymore :-).

PR #66, opened 2023-01-19
Bump versions in GH actions

Bouni:


Bouni:
Sure, I'm aware of that, no problem 🙂

Issue #53, opened 2023-01-16
Add discover function


BenPru:
I have extend bounis discover function and implement it in my integration:
https://github.com/BenPru/luxtronik/blob/3912b8e8219d4c93101ac71df542beeb60b5ed59/custom_components/luxtronik/helpers/lux_helper.py#L11

But I think it should be in the core python luxtronik.
It should discover not only one heatpump in case of more than one. It should return a list with ip or name and port.

Bouni:
You're absolutely right, this should be part of this module rather than the HA integration!
Should be not too hard to move into here, right?

kbabioch:
Having the functionality in the library sounds good. On the other hand I'm not quite sure yet where it would be ideally placed. Probably we need also some kind of helper functions that could be invoked prior to connecting to the controller itself via the `Luxtronik` class.

Bouni:
Maybe if no ip / port is passed when the class is instanciated, it tries to autodiscover the heatpump.
The only questions how it would handle a situation where more then one heatpump is in the same network 🤔

kbabioch:
> Maybe if no ip / port is passed when the class is instanciated, it tries to autodiscover the heatpump.

Personally I'm not in favor of too much "magic" in a Python library. Having a help function that will detect heat pumps and return a list of IPs would be good enough. Automatic connections should be handled by other solutions such as Home Assistant, etc.

As you've said there are some considerations when there are multiple heat pumps in the network for example.

Also it should not be too easy to use the library. What I mean by that is, that (at least to my understanding) there should be some explicit act of commitment, e.g. like specifying an IP and invoking some kind of initialization. If we make it too easy, people could import a module that they don't know yet, and already connect to their heat pump - potentially doing something stupid.

What I think might be a good solution:

- Have functionality to detect heat pumps automatically in the module (some kind of help function). This function will return a list of IPs / ports to connect to.
- Manual initialization (like now)
- Additionally (far future vision): Have a command-line utility (or even GUI) that interacts with the heat pump, i.e. you can interact with all kinds of values from the heat pump via the CLI utility. In this utility we could implement auto-connect features.

BenPru:
Many people install this package manually to get a dump of all values or to find the heatpump.
Why don't we make it as easy as possible for them?
If no param is commited just discover all heatpump and print the ip/name with the port and then a dump of all params and calculations. The visibilities are not relevant I think.
If we detect more than one heatpump just dump the ips and ports.
[...]

PR #70, opened 2023-01-20
Add discover function

Bouni:
This adds the discover function that @BenPru currently has in his HA integration to the core module as suggested in #53.

We might mak this availble from the CLI (see #59) later on.


BenPru (approved):

luxtronik/discover.py
```diff
@@ -0,0 +1,56 @@
+"""Discovery funtion for Luxtronik2 heatpump controllers."""
```
kbabioch, line 1:
```suggestion
"""Discovery function for Luxtronik heat pump controllers"""
```

luxtronik/discover.py
```diff
@@ -0,0 +1,56 @@
...
+import logging
+import socket
+
+LOGGER = logging.getLogger("Luxtronik.Discover")
+
+
+def discover() -> list[(str, int)]:
+    """Broadcast discovery for luxtronik heatpumps."""
```
kbabioch, line 9:
```suggestion
    """Broadcast discovery for Luxtronik heat pumps"""
```

luxtronik/discover.py
```diff
@@ -0,0 +1,56 @@
...
+
+    results: list[(str, int)] = list()
+
+    for magic_port in (4444, 47808):
```
kbabioch, line 13: I would suggest to put those ports into a constant, e.g. something like `LUXTRONIK_DISCOVERY_PORTS`. That way the code is better readable, the values could be re-used and changed later on easily.

BenPru (approved):

luxtronik/discover.py
```diff
@@ -0,0 +1,56 @@
...
+        LOGGER.debug("Send discovery packets to port %s", magic_port)
+        server = socket.socket(socket.AF_INET, socket.SOCK_DGRAM, socket.IPPROTO_UDP)
+        server.setsockopt(socket.SOL_SOCKET, socket.SO_BROADCAST, 1)
+        server.bind(("", magic_port))
+        server.settimeout(2)
```
kbabioch, line 18: The timeout value should be put into a constant also, so it could be better understood. Suggestion: `LUXTRONIK_DISCOVERY_TIMEOUT`

luxtronik/discover.py
```diff
@@ -0,0 +1,56 @@
...
+
+        # send AIT magic broadcast packet
+        data = "2000;111;1;\x00"
```
kbabioch, line 21: This should also be a constant, e.g. `LUXTRONIK_DISCOVERY_MAGIC_PACKET`

luxtronik/discover.py
```diff
@@ -0,0 +1,56 @@
...
+        while True:
+            try:
+                res, con = server.recvfrom(1024)
+                res = res.decode("ascii", errors="ignore")
+                # if we receive what we just sent, continue
+                if res == data:
+                    continue
+                ip_address = con[0]
+                # if the response starts with the magic nonsense
+                if res.startswith("2500;111;"):
```
kbabioch, line 34: This could re-use the constant `LUXTRONIK_DISCOVERY_MAGIC_PACKET`

luxtronik/discover.py
```diff
@@ -0,0 +1,56 @@
...
+                    res = res.split(";")
+                    LOGGER.debug("Received answer from %s %s", ip_address, str(res))
+                    try:
+                        port = int(res[2])
```
kbabioch, line 38: Do we need any kind of sanity check on this value? Maybe we should check whether the int value is in between `1-65535`?

luxtronik/discover.py
```diff
@@ -0,0 +1,56 @@
...
+                    except ValueError:
+                        LOGGER.debug(
+                            "Response did not contain a valid port number,"
+                            "an old Luxtronic software version might be the reason."
+                        )
+                        port = None
+                    results.append((ip_address, port))
```
kbabioch, line 45: In case of `ValueError` in line 39 the port will be `None` at this stage. I guess this is wanted?

luxtronik/discover.py
```diff
@@ -0,0 +1,56 @@
...
+                LOGGER.debug(
+                    "Received answer, but with wrong magic bytes,"
+                    "from %s skip this one",
```
kbabioch, line 48:
```suggestion
                    "Received answer from %s, but with wrong content,"
                    "skipping",
```

kbabioch (commented):
Overall the code itself looks fine. Added a few improvements that could be made.

Also I'm wondering whether this should live in a new/own module (`discover`), or whether it could be a better idea to put it into the `__init__` module along with the `Luxtronik` class.

That way you could use it somewhat like this:

```python
import luxtronik
heatpump = luxtronik.discover()[0]
l = luxtronik.Luxtronik(heatpump[0], heatpump[1])
```

luxtronik/discover.py
BenPru: Yes. IP discovered, port not.

luxtronik/discover.py
```diff
@@ -2,6 +2,8 @@
 import logging
 import socket
 
+LUXTRONIK_DISCOVERY_PORTS = (4444, 47808)
```
kbabioch, line 5: This should be a list from my understanding:

```suggestion
LUXTRONIK_DISCOVERY_PORTS = [4444, 47808]
```

Bouni:
Could we simply import the discover function into __init__.py to be able to use `luxtronik.discover()`?
Just a middle of the night idea 😅

kbabioch:
> Could we simply import the discover function into `__init__.py` to be able to use `luxtronik.discover()`? Just a middle of the night idea sweat_smile

Good idea, probably was too late for me to think about this myself :-).

I've pushed some changes to the branch that addresses this, fixes the linting issue and contains some changes regarding consistency and documentation. Please have a look, otherwise feel free to merge, at least from my point of view it is reviewed :-).

kbabioch (approved):

PR #72, opened 2023-01-20
Change default version in setup.py

Bouni:


kbabioch (approved):
Looks good to me, replace logic should still work.

PR #74, opened 2023-01-23
Improve handling of IP addresses

Bouni:
Replace the use of the `ipaddress` module combined with unclear addition subtraction in favor of `socket` and `struct`.


kbabioch (approved):
Had another look at the changes and the way you address this should work fine with what we currently know about the internal handling of IP addresses. Let's go with it.

PR #76, opened 2023-01-24
Fix wrong data types and flake warnings

BenPru:


kbabioch:
Are all those findings from flake? In general it would be nice to have a (short) description in the pull request, so that we know why this change is wanted/needed.

luxtronik/discover.py
```diff
@@ -1,4 +1,6 @@
-"""Discovery function for Luxtronik heat pump controllers"""
+"""Discovery function for Luxtronik heat pump controllers."""
+from __future__ import annotations
```
kbabioch, line 2: Not sure about this. We seem to be inconsistent with the import of `annotations`. We can go for it, but should we do this (in another pull request) for all modules?

luxtronik/discover.py
```diff
@@ -32,35 +34,37 @@ def discover() -> list[(str, int)]:
...
+                        "Received response from %s %s", ip_address, str(res_list)
+                    )
                     try:
-                        port = int(res[2])
-                        if port < 1 or port > 65535:
-                            LOGGER.debug(
-                                    "Response contained an invalid port, ignoring"
-                                    )
-                            port = None
+                        res_port: int | None = int(res_list[2])
```
kbabioch, line 56: Was this a change from flake8 also? Not sure if this really improves anything, since `port` is also pretty descriptive and shorter than `res_port`.

BenPru: The variable port was double used.
The port to scan and the API port reported by the heatpump answer.
I have created an own variable for the reported port.

luxtronik/discover.py
BenPru: The linter and so in does not allow to assign different types to the same variable.
An int var can not set to None. It has to be declared : int | None.
But this is only possible since Python 3.10.
For <3.10 the annotion import ist needed.

luxtronik/discover.py
```diff
@@ -32,35 +34,37 @@ def discover() -> list[(str, int)]:
...
+                            res_port = None
                     except ValueError:
-                        port = None
-                    if port is None:
+                        res_port = None
+                    if res_port is None:
                         LOGGER.debug(
                             "Response did not contain a valid port number,"
-                            "an old Luxtronic software version might be the reason."
+                            "an old Luxtronic software version might be the reason"
```
BenPru, line 65: W7401: User visible logger messages must not end with a period (hass-logger-period)

luxtronik/discover.py
```diff
@@ -17,10 +19,10 @@
 LOGGER = logging.getLogger("Luxtronik.Discover")
 
 
-def discover() -> list[(str, int)]:
```
BenPru, line 20: "list" expects 1 type argument, but 2 given  [type-arg]

luxtronik/discover.py
```diff
@@ -17,10 +19,10 @@
...
-    """Broadcast discovery for Luxtronik heat pumps"""
+def discover() -> list[tuple[str, int | None]]:
+    """Broadcast discovery for Luxtronik heat pumps."""
 
-    results: list[(str, int)] = []
```
BenPru, line 23: "list" expects 1 type argument, but 2 given  [type-arg]

luxtronik/discover.py
```diff
@@ -32,35 +34,37 @@ def discover() -> list[(str, int)]:
...
         server.sendto(LUXTRONIK_DISCOVERY_MAGIC_PACKET.encode(), ("<broadcast>", port))
-        LOGGER.debug("Sending broadcast request %s", LUXTRONIK_DISCOVERY_MAGIC_PACKET.encode())
+        LOGGER.debug(
+            "Sending broadcast request %s", LUXTRONIK_DISCOVERY_MAGIC_PACKET.encode()
+        )
 
         while True:
             try:
-                res, con = server.recvfrom(1024)
-                res = res.decode("ascii", errors="ignore")
```
BenPru, line 40: mypy: error: Incompatible types in assignment (expression has type "str", variable has type "bytes")  [assignment]
Double use of var res!

luxtronik/discover.py
```diff
@@ -32,35 +34,37 @@ def discover() -> list[(str, int)]:
...
+                recv_bytes, con = server.recvfrom(1024)
+                res = recv_bytes.decode("ascii", errors="ignore")
                 # if we receive what we just sent, continue
                 if res == LUXTRONIK_DISCOVERY_MAGIC_PACKET:
                     continue
                 ip_address = con[0]
                 # if the response starts with the magic nonsense
                 if res.startswith(LUXTRONIK_DISCOVERY_RESPONSE_PREFIX):
-                    res = res.split(";")
```
BenPru, line 47: mypy: error: Incompatible types in assignment (expression has type "List[bytes]", variable has type "bytes")  [assignment]
Double use of var res

BenPru:
> Are all those findings from flake? 

And mypy. I have copied the discover function to my integration for tests and so on. The ha pre-commit-hooks have found this problems.

> In general it would be nice to have a (short) description in the pull request, so that we know why this change is wanted/needed.

I have added the mypy messages.

luxtronik/discover.py
kbabioch: I can understand this, just wondering whether we should apply this to other files as well. But I guess we will eventually add linting via pre-commit, etc. later on to the whole repository anyway (via #77).

kbabioch (approved):

Issue #77, opened 2023-01-25
Introduce pre-commit-hooks


Bouni:
I think we should start using pre-commit-hooks as for example HA does!

See https://github.com/home-assistant/core/blob/dev/.pre-commit-config.yaml for reference

That way we could have nice things like **black** formatting, **pylint** or **ruff** for linting, **isort** for nice import order, and **autoflake** to remove unused imports and variables.
That way we can fix issues before we push code and GH actions show that there are improvements necessary.

Optinions?

kbabioch:
Yes, having pre-commit hooks might add **additional** convenience. Since it can be disabled / ignored locally, we still need the important checks in in the CI/CD pipeline.

Should we just copy the essential stuff from Home Assistant?

Bouni:
Yes, that was my intention. I work on it right now, but I think I went down a deep rabbit hole 🙈 

- setup.py -> legacy, should be replaced with pyproject.toml
- pyproject.toml -> looking at HA's pyproject.toml, jeez what are all these settings 🤯 

A lot to cleanup I guess, maybe we should tackle that in a first PR and after that's done start with the pre-commit-hooks

kbabioch:
@Bouni Any help you need here? I guess you're working on this already in the background and have already prepared some changes? Looking forward to those changes ;-).

Bouni:
I worked on this last week. The biggest issue is that I need to change many thing in order to get the pre-commit hooks to be statisfied 😆 

I tink I'll add them commented out in the first place, then we can "activate them" one by one and change the necessary files. Otherwise we end up with a giant change set.

One more thing, I would really like to add `black` as the first pre-commit hook becuase that will solve a lot of ruff/pylint warnings/errors because it wraps long lines for example.

Are you ok with that?

kbabioch:
Yes, for me that is totally fine and having black as pre-commit hook is also okay.

However I would not like to have automatic rewritting of code as part of the pipeline, i.e. we could run black only in some kind of reporting mode when we integrate it into the pipeline.

The "problem" with pre-commit is, that its more of a "nice-to-have", and not mandatory. So not everyone will execute it when submitting PRs.

Bouni:
We could run black with the `--check` flag: https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#writeback-and-reporting

but to be honest, I run black on almost all my projects and it never produced code that didn't work afterwards.
[...]

PR #78, opened 2023-01-25
Replace setup.py with pyproject.toml

Bouni:
As mentioned in #77 I realized that `setup.py` is kind of deprecated and its recommended to switch to a `pyproject.toml` instead.

I did that with this PR and already added optional dev dependencies that will be used in the pre-commit-hooks PR that is upcoming. @kbabioch @BenPru If you want, I'll add you to the list of contributors as well!


pyproject.toml
```diff
@@ -1,6 +1,57 @@
...
+  "flake8-docstrings~=1.6.0",
+  "flake8-noqa~=1.3.0",
+  "flake8~=6.0.0",
+  "isort~=5.11.4",
+  "pycodestyle~=2.10.0",
+  "pydocstyle~=6.2.3",
+  "pyflakes~=3.0.1",
+  "pyupgrade~=3.3.1",
+  "ruff~=0.0.231",
+  "yamllint~=1.28.0"
```
kbabioch, line 48: Do we need the version pinning here? How are we going to manage this? How often are we supposed to bump those versions?

kbabioch (approved):

pyproject.toml
Bouni: Those are shamelessly stolen out of HA's pyproject.toml 😇 
But I guess you're right, better not pin them.

kbabioch: Well, version pinning is a hot topic and you could argue back and forth all time long. I just wonder how we're supposed to manage those numbers. I don't know all of those libraries/modules/tools, so I have no idea, which version to use. Sticking with old version for forever is probably also not a good idea.

We could add a reference to the Home Assistant file and check/compare it every now and then. I'm sure someone in the Home Assistant project takes care about such things :-).

PR #79, opened 2023-01-26
Replace pylint with ruff in GH Actions

Bouni:
This morning I heared the Epsiode 400 of the [Talk Python Podcast](https://talkpython.fm/episodes/show/400/ruff-the-fast-rust-based-python-linter) in which they spoke about `ruff` a new linter written in Rust that  10-110 times faster than other liters.
So I thought wh not use it instead of pylint which is the slowest in their comparison:

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -28,6 +28,22 @@ jobs:
         run: pip install --user yamllint
       - name: Run yamllint
         run: yamllint .
+  pylint:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.9", "3.10", "3.11"]
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install pylint
+        run: pip install pylint
+      - name: Run pylint
+        run: pylint $(git ls-files '*.py')
   ruff:
     runs-on: ubuntu-latest
     strategy:
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -23,12 +23,12 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@v4
         with:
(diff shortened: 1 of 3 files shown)


kbabioch:
I'm not sure if "faster" in our project / use case is a relevant argument (run time < 1 second or so), but let's go with it (it was added with #77) anyway already.

Bouni:
>personally I don't think that 88 characters is a good line length in 2023

Totally agree! 120 seems like a sane value to me.

I added the # noqa: F401 to the import of the dicovery function in `__init__.py` so that we can use `luxtronik.discover()` instead of `luxtronik.discover.discover()`

kbabioch:
We still have some `# pylint` comments in our code base. Should we remove those, since pylint will no longer be used?

Bouni:
I think we might have to adjust them to ruff comments, maybe ruff reacts to pylint comments as well 🤔

Bouni:
I removed all `# pylint:` lines and ruff does not complain about them.
The reason is that not all pylint rules are yet implemented: https://github.com/charliermarsh/ruff/issues/970

kbabioch:
> I removed all `# pylint:` lines and ruff does not complain about them. The reason is that not all pylint rules are yet implemented: [charliermarsh/ruff#970](https://github.com/charliermarsh/ruff/issues/970)

So when not all rules are implemented, is it still a good idea to use it already **instead** of `pylint`. Maybe we should add it **in addition** and switch once it is "feature complete". We're not paying for runtime here, so running two linters doesn't hurt.

Otherwise we might need to fix the code later on ;-).

Bouni:
Well, you got a point there. I'll add ruff additionaly to pylint!

.github/workflows/lint.yml
```diff
@@ -23,12 +23,12 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@v4
         with:
-          python-version: '3.11'
+          python-version: "3.11"
```
kbabioch, line 26: The change itself is fine, although the commit message is unrelated to the change. It's trivial and not important, but we should try to avoid this in the future by having good commit messages ;-).

kbabioch (approved):

PR #81, opened 2023-01-30
Add pre-commit-hooks

Bouni:
This is meant to be a starting point. We'll add the hooks one after another to prevent to many changes to be made at a time.

Here we start with `black` as  the first pre-commit-hook.

make sure you have installed pre-commit (` pip install pre-commit`) then simply run `pre-commit install` and you set.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,75 +1,70 @@
 repos:
-  - repo: https://github.com/charliermarsh/ruff-pre-commit
-    rev: v0.0.235
-    hooks:
-      - id: ruff
-        args:
-          - --fix
-  - repo: https://github.com/asottile/pyupgrade
-    rev: v3.3.1
-    hooks:
-      - id: pyupgrade
-        args: [--py310-plus]
-        stages: [manual]
-  - repo: https://github.com/PyCQA/autoflake
-    rev: v2.0.0
-    hooks:
-      - id: autoflake
-        args:
-          - --in-place
-          - --remove-all-unused-imports
-        stages: [manual]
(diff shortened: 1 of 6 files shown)


.pre-commit-config.yaml
```diff
@@ -0,0 +1,70 @@
+repos:
```
kbabioch, line 1: We should add `---` at the beginning of this document in order to address the issue:

1:1 [document-start] missing document start "---"

.pre-commit-config.yaml
```diff
@@ -0,0 +1,70 @@
...
+  - repo: https://github.com/psf/black
+    rev: 22.12.0
+    hooks:
+      - id: black
+        files: ^((luxtronik|pylint|scripts|tests)/.+)?[^/]+\.py$
+  # - repo: https://github.com/charliermarsh/ruff-pre-commit
```
kbabioch, line 7: > 7:3 [comments-indentation] comment not indented like content

kbabioch (approved):

kbabioch:
One more thing: This file is ending in `.yaml`, other YAML files in our repository end in `.yml`. Should we try to stay consistent here?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,71 @@
+---
```
kbabioch, line 1: One more thing: This file is ending in `.yaml`, other YAML files in our repository end in `.yml`. Should we try to stay consistent here?

Bouni: I just tried and pre-commit expects a .yaml file, .yml is not recognized 🤷🏽‍♂️

PR #82, opened 2023-01-30
Add black in RO mode to lint GH action

Bouni:
As discussed in #77 I've added `black` to the lint CI job.

Info: https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#writeback-and-reporting

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -30,32 +30,26 @@ jobs:
         run: yamllint .
   black:
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        python-version: ["3.9", "3.10", "3.11"]
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
       - name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v4
         with:
-          python-version: ${{ matrix.python-version }}
+          python-version: "3.11"
       - name: Install black
         run: pip install black
       - name: Run black
         run: black --check --verbose $(git ls-files '*.py')
   pylint:
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        python-version: ["3.9", "3.10", "3.11"]
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


Bouni:
@kbabioch Is there a reason for installying pytest within the pylint job?: https://github.com/Bouni/python-luxtronik/blob/00b3ee8dedade655d8b349b371ee4ab164c5ebd3/.github/workflows/lint.yml#L59-L60

kbabioch:
> @kbabioch Is there a reason for installying pytest within the pylint job?:
> 
> https://github.com/Bouni/python-luxtronik/blob/00b3ee8dedade655d8b349b371ee4ab164c5ebd3/.github/workflows/lint.yml#L59-L60

Without it, `pylint` would complain about an import error, see: 8dfeffb90b19a24c5a13a0476de359338e19730c

.github/workflows/lint.yml
```diff
@@ -28,6 +28,22 @@ jobs:
         run: pip install --user yamllint
       - name: Run yamllint
         run: yamllint .
+  black:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.9", "3.10", "3.11"]
```
kbabioch, line 35: Not sure, do we really need to run this with multiple versions of Python?

Bouni: I don't think so, in my opinion having only 3.11 is just fine!

Bouni: @kbabioch Any thoughts on my last question about pylint?

kbabioch: I'm fine with only having 3.11 for linting as well. Sorry for the late reply, somehow I ran into "out of time" errors ;-).

Bouni:
@kbabioch Do you know, is there a way to run actions just once and not for push and pull_request on every commit?

kbabioch:
> @kbabioch Do you know, is there a way to run actions just once and not for push and pull_request on every commit?

No, unfortunately that seems to be not totally trivial. Had looked into this some weeks ago, but couldn't figure it out quickly.

PR #87, opened 2023-03-09
Add more linter rules

Bouni:
I realized that the PR #86  does fail because of pylint, so I added some sane pylint configs to pyproject.toml

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -56,5 +56,22 @@ include-package-data = true
 [tool.setuptools.packages.find]
 include = ["luxtronik*"]
 
+[tool.pylint.format]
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -56,5 +56,22 @@ include-package-data = true
 [tool.setuptools.packages.find]
 include = ["luxtronik*"]
 
+[tool.pylint.format]
+max-line-length = 120 # Set a better lin length limit, 80 chars is just archaic
```
kbabioch, line 60:
```suggestion
max-line-length = 120 # Set a better line length limit, 80 chars is just archaic
```

PR #94, opened 2023-03-31
issue-93: Also output the suggested changes of the 'black' workflow

### LINKED ISSUE #93: Python Black workflow output insufficient
The "black" workflow only states that some files should be changed, but does not report these changes.
Adding a `--diff ` option to the call would fix this.

Guzz-T:
The "black" workflow only states that some files should be changed, but does not report these changes.
Fixed this by adding a --diff option to the call.

Fixes #93

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -40,7 +40,7 @@ jobs:
       - name: Install black
         run: pip install black
       - name: Run black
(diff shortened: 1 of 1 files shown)


kbabioch (approved):

Bouni:
LGTM

Issue #49, opened 2023-01-16
datatype value is actually not used


kbabioch:
The base class for all datatypes is `Base`, which looks as follows:

https://github.com/Bouni/python-luxtronik/blob/34f4e666bddd0c380fd0781de76a216d5708843b/luxtronik/datatypes.py#L6-L28

All other datatypes are derived from that. I'm currently wondering what the attribute `value` is actually used for in this context. The `__repr__` and `__str__` methods are referring to it, but actually the value is never set (other than during initialization). Any invocation of `from_heatpump` and `to_heatpump` will just calculate and return the value, but never set it.

Is this attribute actually needed and if so, what for and how is it used? Am I missing something here?

Bouni:
I just had to look what I've been doing back then 😅 

When all the data is read from the heatpump, we pass the raw value to `from_heatpump` and store the result in `.value`, here for example:

https://github.com/Bouni/python-luxtronik/blob/main/luxtronik/parameters.py#L1171

When you then get a parameter (`tvl = l.calculations.get("ID_WEB_Temperatur_TVL")`) you use it either like `print(tvl)` which uses `__str__` or like `print(tvl.value)`

As i read the code and think about it, not very nice what I did there :woozy_face:

Maybe we should read all data, store it in `.raw_data`, and convert it on the fly:

```python

class Base:
    """Base datatype, no conversions."""

    measurement_type = None

    def __init__(self, name, writeable=False):
        self.raw_value = None
        self.name = name
        self.writeable = writeable

    def _to_heatpump(self):
        """Converts value into heatpump units."""
        return self.raw_value

    def _from_heatpump(self):
        """Converts value from heatpump units."""
        return raw_value

    @raw_value.setter
    def set_raw_value(self, value):
        self.raw_value = self._to_heatpump(value)
        return self.raw_value

    @property
    def value(self):
        return self._from_heatpump(self.raw_value)

    def __repr__(self):
        return str(self.value)

    def __str__(self):
        return str(self.value)
```

That way we could have `print(tvl)` as well as `print(tvl.value)` and `print(tvl.raw_value)`, and for writing `self.queue[index] = parameter.value = 12345` (https://github.com/Bouni/python-luxtronik/blob/main/luxtronik/parameters.py#L1210)

These are just thoughts, nothing tested ...

kbabioch:
In general I like your approach, that might turn out useful.

[...]

PR #96, opened 2023-05-01
issue-49: Convert some datatype functions into class methods

Guzz-T:
As discussed in #49 the implementation of the functions `Base.to_heatpump`, `Base.from_heatpump` and `SelectionBase.options` are correspond to class methods. But are not yet offered as such.  This pull-request transform them into class-methods.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,7 +107,7 @@ l.write(parameters)
 
 # If you're not sure what values to write, you can get all available options:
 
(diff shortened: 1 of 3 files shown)


kbabioch:
For now we don't seem to have agreed on an overall architecture, so I'm not entirely sure what "the best" way to handle those methods is. I'm fine with the change, maybe will have to re-factor it in the future further.

Issue #35, opened 2023-01-11
Understanding what the parameters/calculations mean


pmatos:
Hi,

I have a few questions:

1. What's the difference between parameters and calculations?
2. Do you hold a file somewhere about the meaning of the parameters you know about?

For example, If I open my Alpha Innotec Control application I can see values and explanations, I was wondering if I could print all values available and try to associate them to an explanation of what they are. I could then add to the documentation you already publicly have. What do you think?

Also, is there a way to cycle through all parameters, calculations and print their values?

kbabioch:
I'm familiar with gettext and have used it across some projects (although not necessarily with Python). 

However I'm not yet convinced that "abusing" strings as part of the class is the right way to document the parameters. While something like `Temperatur Vorlauf` is already more descriptive than `ID_WEB_Temperatur_TVL`.

In a perfect world the documentation should also contain more details, i.e. what this parameter means (physically), what the allowed values are (data type, min, max, etc.).

Typically I would suggest to use docstrings for this, so that the documentation lives next to the code to ensure it stays consistent as good as possible.

There are some problems though:

- The essential aspects that need documentation are entries in a dictionary. Not sure how well docstrings will work for that, some initial tests were not too promising.
- No i18n: The code will be documented in English most likely. I'm not familiar with docstrings in multiple languages. For me English is perfectly fine, but some users of the library might be unhappy with that. I'm pretty sure that the user base of this library are owners of AIT products, which is probably widespread in Germany, but nowhere else?

Bouni:
Maybe the docs could live in a seperate file and matched by the id, that way we could have a proper name, description, min, max etc. as well as i18n support (at least thats waht I assume, nver really used it)

Something like 

**docs.py**
```python
paramter_docs = {
  10: {
    "name": _("Vorlauftemperatur"),
    "description": _("bla bla"),
    "min": 10.0,
    "max": 50.0
  }
}
```

`_()` the is used to get the correct translation via `gettext`.

I guess we need to pass the id then into the class to have it available for getting the docs, something like `10: Celsius("ID_WEB_Temperatur_TVL", 10),`

BenPru:
> **docs.py**
> ```python
> paramter_docs = {
> ...
>   10: {
>     "name": _("Vorlauftemperatur"),
>     "description": _("bla bla"),
>     "min": 10.0,
>     "max": 50.0
>   },
> ...
> }
> ```

[...]

PR #101, opened 2023-07-07
Details for two parameters related to hot water.

gerw:
This patch adds details for:

* Parameter 994. This parameter decides if the heating rod is enabled to heat the hot water if the heat pump does not reach the desired temperature (Parameter 105) or if the temperature reaches the value of parameter 2.
* Parameter 1020. I think that this parameter controls, how long the heating rod is allowed to heat the water.

Both parameters can be accessed at a German heat pump via "Service" -> "Einstellungen" -> "System Einstellungen" -> "Warmw. Nachheizung" / "Warmw. Nachh. max"

Parameter 1020 uses a new conversion formula for hours. ["Fun" fact: if you register as "Kundendienst" / "after sales service", then the parameter value is displayed (with `raw/10`) incorrectly in the "Event Log".]

By the way: I think that there was some discussion, whether more information concerning the parameters could be collected in this module. My suggestion would be to move the parameter information to a separate (json formatted?) file, which is consequently read by `parameters.py`. Then, one could also add something like max and min values for the parameters. Would there be any downside?


kbabioch (approved):

kbabioch:
Changes look fine to me. Personally I would also like to see unit tests, as they are already available for the other datatypes.

Regarding the json file for meta information: I like the idea, but would encourage you to discuss is within #35 rather than in this pull request.

gerw:
I added a simple unit test.

Bouni (approved):
Lgtm

gerw:
Since the PR was not merged yet, I took the opportunity to identify some more calculations. For some reason, the energy from the heating rod seems to be a parameter.

Calculations 232,233, 236-243 can be seen in the web interface if you are logged in as after-sales service. I also have taken the liberty to rename calculation 241, since "Circulation_Pump" sounds (at least for me) like Zirkulationspump/ZIP which is something different.

For me, calculations 234 and 235 are always "0".

By the way: My heating rod uses 6kW but is configured to 9kW. This seems to be parameter `1025` with--yet another--a scaling factor of `100W`. Can anybody confirm this? However, I do not find a menu in which I could set my heating rod to 6kW.

luxtronik/calculations.py
```diff
@@ -264,18 +264,18 @@ def __init__(self):
...
-            242: Unknown("Unknown_Calculation_242"),
-            243: Unknown("Unknown_Calculation_243"),
+            236: Frequency("ID_WEB_Freq_VD_Soll"),
+            237: Frequency("ID_WEB_Freq_VD_Min"),
+            238: Frequency("ID_WEB_Freq_VD_Max"),
+            239: Kelvin("VBO_Temp_Spread_Soll"),
+            240: Kelvin("VBO_Temp_Spread_Ist"),
+            241: Percent2("HUP_PWM"),
+            242: Kelvin("HUP_Temp_Spread_Soll"),
+            243: Kelvin("HUP_Temp_Spread_Ist"),
```
kbabioch, line 278: Wondering what this means physically? Don't remember to have seen this in the menu (will check later). But `HUP` means `Heizungsumwälzpumpe`. To my knowledge it is just circulating the water in the heating system. Not entirely really sure what `Soll` and `Ist` are supposed to mean in this context?

How exactly are those parameters called in the menu?

kbabioch:
> By the way: My heating rod uses 6kW but is configured to 9kW. This seems to be parameter `1025` with--yet another--a scaling factor of `100W`. Can anybody confirm this? However, I do not find a menu in which I could set my heating rod to 6kW.

That's interesting. For me it says:

> Number: 1025  Name: ID_Einst_Leistung_ZWE                                        Type: Unknown              Value: 90

As you've said that could mean `9000` with a scaling factor of `100`. Physically it's only 3 kW in my case, and I never seen a menu to change this. Would be interesting to learn what this is used for. Maybe some internal calculations?

kbabioch:
In general: Thank you for working on identifying more parameters. That's appreciated.

I think we should try to avoid to have too many of those parameters introduced / changed in one MR, though in the future. Ideally, for every set of datapoints that are related to each other, there should be a dedicated MR.

Otherwise the discussion will get more and more confusing and it's hard to follow along and finish something.

So for the future, I would suggest to create a new MR for new datatypes.

luxtronik/calculations.py
```diff
@@ -264,18 +264,18 @@ def __init__(self):
...
-            239: Unknown("Unknown_Calculation_239"),
-            240: Unknown("Unknown_Calculation_240"),
-            241: Percent2("Circulation_Pump"),
-            242: Unknown("Unknown_Calculation_242"),
-            243: Unknown("Unknown_Calculation_243"),
+            236: Frequency("ID_WEB_Freq_VD_Soll"),
+            237: Frequency("ID_WEB_Freq_VD_Min"),
+            238: Frequency("ID_WEB_Freq_VD_Max"),
+            239: Kelvin("VBO_Temp_Spread_Soll"),
+            240: Kelvin("VBO_Temp_Spread_Ist"),
```
kbabioch, line 275: What does `VBO` mean in this context?

luxtronik/calculations.py
gerw: These values can be seen in the (new) webinterface of the heat pump if you are logged in as Kundendienst.

If I understand my heat pump correctly, it only has two pumps: One for pumping the water into the bore holes (this is `VBO` which seems to come from `Ventilator, Brunnen- oder Soleumwälzpumpe`) and another one for pumping the heated water either into the hot water deposit or into the underfloor heating pipes).

These values report the `Ist` and `Soll` values of the temperatures of inflow and outflow. I think they are used to control the operating level of the pump.

luxtronik/calculations.py
gerw: See above.

gerw:
> > By the way: My heating rod uses 6kW but is configured to 9kW. This seems to be parameter `1025` with--yet another--a scaling factor of `100W`. Can anybody confirm this? However, I do not find a menu in which I could set my heating rod to 6kW.
> 
> That's interesting. For me it says:
> 
> > Number: 1025  Name: ID_Einst_Leistung_ZWE                                        Type: Unknown              Value: 90
> 
> As you've said that could mean `9000` with a scaling factor of `100`. Physically it's only 3 kW in my case, and I never seen a menu to change this. Would be interesting to learn what this is used for. Maybe some internal calculations?

In the webinterface, one can see the amout of heat produced by the heating rod. This seems to be also Parameter 1059. That is, parameter 1059 increases when the heating rod is on, and for this, parameter 1025 is used.

gerw:
Concerning multiple MR: Yes, I agree.

kbabioch:
> Concerning multiple MR: Yes, I agree.

To speed things up: Could you split your work, so that we stick to the first two commits (already approved) in this MR?

Something like this should do the trick:

> git checkout main
> git checkout -b new-parameters
> git checkout main
> git reset --hard  f47c37b996cf6bae0fa3036a09ee1382484fc2b0
> git push --force

Afterwards your main branch (source for this MR) will be in the previous state and you'll have your other commit in a `new-parameters` branch.

Bouni:
@kbabioch Feel free to merge this one as you have looked into this a lot more than I did 😅

kbabioch:
There is one more minor issue. The linting now complains about:

https://github.com/Bouni/python-luxtronik/actions/runs/5520635370/jobs/10068627802?pr=101#step:6:13

> luxtronik/datatypes.py:281:4: W0221: Number of parameters was 2 in 'Base.from_heatpump' and is now 2 in overriding 'Hours2.from_heatpump' method (arguments-differ)
luxtronik/datatypes.py:284:4: W0221: Number of parameters was 2 in 'Base.to_heatpump' and is now 2 in overriding 'Hours2.to_heatpump' method (arguments-differ)

The reason is the refactoring of the classes via #96. Unfortunately I don't have permission to change your branch, so could you refactor it on top of current `main`? Essentially you'll have to add the `@classmethod` keyword.

kbabioch:
> @kbabioch You mean rebase on main, right? not refactor!?

Yes, exactly. Sorry for confusion. It needs to be rebased on top of current `main`.

kbabioch:
Changes look good now, merging it into `main`. Thank you for the contribution. Let's discuss the other parameters in follow-up pull requests.

kbabioch:
> @kbabioch : What about my last comment of `Hours2.to_heatpump` vs `Hours2("").to_heatpump`?

Hm, seem to have missed that.

Can it be removed, though? `to_heatpump` is a class method, but we are still dealing with an instance of an object, aren't we?

kbabioch:
Addressed with #105. Also identified an issue identified by closely looking at the CI/CD output (see #104). Need to address the issues with unit tests with `SelectionBase` now.

PR #102, opened 2023-07-11
Introduce constants to use for magic values in the Luxtronik class

kbabioch:
This establishes constants for several magic values that are used within the
Luxtronik class. This makes it easier to understand the purpose of these
values and will simplify any kind of refactoring later on.


Bouni:
Maybe its a good idea to put all these into a seperate `const.py` file!?

PR #105, opened 2023-07-11
tests: Don't invoke class methods on class instances

kbabioch:
This adjusts test cases for datatypes in such a way that the recently introduced class methods are invoked on the class rather than on an class instances.


Bouni (approved):
Lgtm!

PR #106, opened 2023-07-11
pytest: Enable pipefail

### LINKED ISSUE #104: pytest errors are being ignored
Errors reported by pytest as part of the GitHub action are currently being ignored, see: https://github.com/Bouni/python-luxtronik/actions/runs/4854647237/jobs/8652268562

The root cause is the pipe into `tee`, which suppresses the error. This could be addressed with `set -o pipefail`.

kbabioch:
This enables the pipefail option as part of the pytest invocation in order to not suppress errors detected by pytest due to the piping into `tee` afterwards.

This addresses and closes #104.

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -16,10 +16,11 @@ jobs:
       - name: Install pytest
         run: pip install pytest pytest-cov
       - name: Run pytest
-        run: >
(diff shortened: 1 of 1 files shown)


kbabioch:
> Lgtm! Just out of curiosity, wasn't the multiline version working?

No, somehow it was broken, since I have to issue multiple commands now (rather than a single command).

PR #108, opened 2023-07-11
Fix unit tests for SelectionBase

kbabioch:
This addresses issuees with the unit tests of SelectionBase, as they have not been working since the usage of class methods. In order to fix those unit tests a new child class of SelectionBase is required, which sets some codes that are then being evaluated by the test methods, since any class methods don't have access to codes that are being set on a per-instance level.


kbabioch:
This addresses and closes #108.

Bouni (approved):
Lgtm :+1:

PR #109, opened 2023-07-11
Some more calculations

gerw:
This was already mentioned in #101:

I took the opportunity to identify some more calculations. For some reason, the energy from the heating rod seems to be a parameter.

Calculations 232,233, 236-243 can be seen in the web interface if you are logged in as after-sales service. I also have taken the liberty to rename calculation 241, since "Circulation_Pump" sounds (at least for me) like Zirkulationspump/ZIP which is something different.

For me, calculations 234 and 235 are always "0".

By the way: My heating rod uses 6kW but is configured to 9kW. This seems to be parameter 1025 with--yet another--a scaling factor of 100W. Can anybody confirm this? However, I do not find a menu in which I could set my heating rod to 6kW.


kbabioch (approved):

kbabioch:
Have you tried (or does it even make sense) to write parameter 1025 in order to represent the actual power that is used / attached to the heating element? If this parameter is used internally to integrate the energy consumed, it will be wrong by quite a bit (33% or 66%), depending on the actual physical connection.

In my case I have only 3 kW attached, so I'm wondering whether I can easily change this value? As you've already mentioned there seems to be no such setting in the menues, etc.?

Maybe we should contact the support about this :-)? Last year they have been quite friendly towards me, but haven't been in touch for quite a while.

gerw:
No, I did not write to parameter 1025. Since I am not sure what this parameter represents, I do not feel like writing to it...

And yes, the energy of the heating rod is integrated incorrectly (by using 9kW)

kbabioch:
> No, I did not write to parameter 1025. Since I am not sure what this parameter represents, I do not feel like writing to it...
> 
> And yes, the energy of the heating rod is integrated incorrectly (by using 9kW)

Okay, thank you for the heads up. Let's see if I feel brave enough to write this value at some point. Doesn't sound too dangouers, but at the end there is no guarantee of course.

kbabioch:
> This seems to be parameter 1025 with--yet another--a scaling factor of 100W. Can anybody confirm this? However, I do not find a menu in which I could set my heating rod to 6kW.

Will you prepare another pull request for this parameter? As said in the other discussion, I can confirm that his is 90 for me, although physically I only have 3 kW connected.

PR #111, opened 2023-07-11
Add low-level helper routines for data exchange

gerw:
This PR tries to remove some duplicated code by introducing low-level helper functions for receiving and sending data.

BTW: I am wondering about the `except struct.error as err`: In which circumstances can the `struct.unpack` raise an exception? I think only if it gets not the correct number of bytes. This, in turn, can only happen, if the `_socket.recv` does not return the correct number of bytes (which is, by documentation, possible). Maybe this should be handled in the new reading routines?


kbabioch:
> BTW: I am wondering about the except struct.error as err: In which circumstances can the struct.unpack raise an exception? I think only if it gets not the correct number of bytes. This, in turn, can only happen, if the _socket.recv does not return the correct number of bytes (which is, by documentation, possible). Maybe this should be handled in the new reading routines?

Are you suggesting to add this in this pull request, or will you prepare this later on? Should we create an issue for this?

luxtronik/__init__.py
```diff
@@ -210,26 +181,34 @@ def _read_calculations(self):
...
             except struct.error as err:
                 # not logging this as error as it would be logged on every read cycle
                 LOGGER.debug("%s: %s", self._host, err)
         LOGGER.info("%s: Read %d visibilities", self._host, length)
         visibilities = Visibilities()
         visibilities.parse(data)
         return visibilities
+
+    def _send_ints(self, *ints):
+        "Low-level helper to send a tuple of ints"
```
kbabioch, line 201:
```suggestion
        "Low-level helper to send list of ints"
```

kbabioch (approved):

kbabioch (approved):

kbabioch:
Deleting more lines that are re-added. I like those refactorings. Makes the code better understandable :-).

PR #116, opened 2023-07-13
Further datatype improvements

kbabioch:
Another set of improvements to the datatype `Base` class, namely:

- Adding docstrings to all methods of the `Base` class
- Checking for instance in `__eq__` method to prevent errors when comparing against objects that are not inherited from the `Base` class


gerw:
Lgtm.

Bouni (approved):
Nice, LGTM!

Issue #99, opened 2023-06-28
Calculation ID 258


RobinFlikkema:
Hi all,

It seems that ID 258 in calculation.py is returning the software version running on the RBE 

For me this is 1.12 and the data returned is "112". My luxtronik controller runs version v3.85.8

Anyone else seeing the same?

kbabioch:
What exactly is RBE in this context?

Bouni:
I think it's short for "RaumBedienEinheit", a remote control thingy for the heat pump

kbabioch:
Ah, okay, don't have it, so can't verify.

Bouni:
@RobinFlikkema I do not have a RBE but your guess seems right.
Any chances that you update the RBE Firmware sometime soon so that we get a chance to see if the number changes?

RobinFlikkema:
I would try that if I knew how to update that thing.

Maybe someone else has this RBE with a different firmware and can crosscheck?

Op 29 jun. 2023 16:06 schreef bouni ***@***.***>:

@RobinFlikkema<https://github.com/RobinFlikkema> I do not have a RBE but your guess seems right.
Any chances that you update the RBE Firmware sometime soon so that we get a chance to see if the number changes?

kbabioch:
For me this value is:

> Number: 258   Name: Unknown_Calculation_258                                      Type: Unknown              Value: 0

Probably 0 means "no RBE" then. Are we sure that only value 258 is relevant here, i.e. that the version is represented in only on parameter and not i.e. in multiple ones? What is 259 saying for you?

In my case it is:

> Number: 258   Name: Unknown_Calculation_258                                      Type: Unknown              Value: 0
> Number: 259   Name: Unknown_Calculation_259                                      Type: Unknown              Value: 1

RobinFlikkema:
Hi @kbabioch 

See here:
> Value 256: 0
> Value 257: 0
> Value 258: 112
> Value 259: 1

So unsure if there is a connection with 259 for example

kbabioch:
Do you have a screenshot from the web interface showing this value, i.e. how exactly it is called?

RobinFlikkema:
![image](https://github.com/Bouni/python-luxtronik/assets/4777401/e7074a3e-f8bd-440b-9133-b00c24196e37)
It is shown here at the bottom, freely translated "Firmware version RBE"

kbabioch:
[...]

PR #118, opened 2023-07-13
Introduce datatype MajorMinorVersion to handle calculation 258

kbabioch:
This introduced a new datatype (`MajorMinorVersion`) that is used to interpret calculation 258. It has been reported (see #99) that this calculation is used to encode the version information from the "Raumbedieneinheit" (RBE).

For me this value is 0 (don't have this hardware), for the reporter of #99 it is `112`, which is being shown as `1.12` in the web interface.

The pull request also contains test cases that cover the currently known values.

### DIFF
--- a/luxtronik/datatypes.py
+++ b/luxtronik/datatypes.py
@@ -388,7 +388,7 @@ class MajorMinorVersion(Base):
         if value > 0:
             major = value // 100
             minor = value % 100
-            return ".".join(str(x) for x in [major, minor])
+            return f"{major}.{minor}"
         return "0"
 
 
--- a/luxtronik/datatypes.py
+++ b/luxtronik/datatypes.py
@@ -378,6 +378,20 @@ class Version(Base):
         return "".join([chr(c) for c in value]).strip("\x00")
 
 
+class MajorMinorVersion(Base):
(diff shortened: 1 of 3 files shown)


gerw:
Lgtm.

luxtronik/datatypes.py
```diff
@@ -378,6 +378,20 @@ def from_heatpump(cls, value):
...
+    """MajorMinorVersion datatype, converts from and to a RBEVersion"""
+
+    datatype_class = "version"
+
+    @classmethod
+    def from_heatpump(cls, value):
+        if value > 0:
+            major = value // 100
+            minor = value % 100
+            return ".".join(str(x) for x in [major, minor])
```
Bouni, line 391: Is there any advantage over this vs using an f-string?

I find f-string much more readable, something like `f"{major}.{minor}"`

kbabioch: Nah, you're right. There is absolutely no advantage in my more complicated approach, somehow just didn't have f strings at mind. Was too much inspired by the code from `Version` and haven't given it another thought. Addressed with 95df02d.

Bouni (approved):

PR #120, opened 2023-07-14
Add yamllint and markdownlint to pre-commit-hooks

Bouni:
In order to get a consistent quality of our README.md (and probably other Markdown files to come) we should add markdownlint to our pre-commit-hooks.  
Yamllint was added to ensure the .pre-commit-hooks.yaml is formated correctly as well.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,6 +5,18 @@ repos:
     hooks:
       - id: black
         files: ^((luxtronik|pylint|scripts|tests)/.+)?[^/]+\.py$
+
+  - repo: https://github.com/igorshubovych/markdownlint-cli
+    rev: v0.35.0
+    hooks:
+      - id: markdownlint
+        args:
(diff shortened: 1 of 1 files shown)


kbabioch:
Do we need to specify which files are to be scanned? At least for the GitHub action it is specified more explicitly:

- `markdownlint "**/*.md"`
- `yamllint .`

Bouni:
I've tested locally and without specifying a glob pattern it works perfectly fine for `README.md` and `.pre-commit-hooks.yaml`

kbabioch:
My concern is more that there might be a difference in which files will be evaluated. Right now we're dealing with all files in the GitHub action. If we need to ignore a specific file / directory, I don't know how to do it with the `pre-commit` file, since there is no glob, etc.

But since we're scanning everything for now, let's worry about this, once we're facing this situation ;-).

kbabioch (approved):

Issue #122, opened 2023-07-16
Create base class for Calculations, Parameters, Visibilities


gerw:
These three classes share a good amount of code. There should be a common base class from which these three are inherited.

Issue #119, opened 2023-07-13
Provide high level interface


kbabioch:
Currently this module focuses on providing low level data as reported back from the heat pump itself. While there are different categories of data (parameters, visibilities and calculations), they are essentially reported back as-is with only a simple mapping that is done by the available datatypes. For instance values like `0`, `1`, `2` might be mapped to some string representation:

```
    codes = {
        0: "Automatic",
        1: "Second heatsource",
        2: "Party",
        3: "Holidays",
        4: "Off",
    }
```

As it turns out there are some issues with this, though:

For instance, `Betriebszustand` reports `EVU`, while the heat pump is not allowed to run, while it doesn`t report `cooling` when it is actually cooling without the compressor running (see #112). There are more such glitches discussed / reported in #84.

For now @BenPru has dealt with these complexities in this Home Assistant integration. However it would be nice if those could actually be handled by this module, so that consumers of this module don't have to implement this on their own.

For this a high level interface is needed that represents the current state of the heat pump by interpreting the low level values itself.

Besides that the high level interface could also provide services / functions to the user, so that the user of the module does not need to deal with the low level details.

This could include:

- Setting the room temperature (admittedly this is essentially writing only a single value, but still its an improvement if you don't need to know anything about parameters, etc.)
- Enabling the heating rod to heat up some water (e.g. this could be used to get rid of excess power produced by PV)
- Turn on the `HUP` (Heizungsumwälzpumpe) to get some water flow in the heating system
- Turn on the `ZIP` (Zirkulationspumpe) to get some water flow in the water installation (e.g. when presense is detected by a motion sensor, etc.)

The exact details and additional functionality that might be needed/useful can be discussed in this issue, since I'm interested in the point of view from other users of this module also.

@gerw @Bouni @BenPru What are your thoughts about this?

gerw:
In principle, this is a good idea. But (of course) the user has to take care, that the things are also turned off later. The heat pump itself is not able to do it.

I am also developing (or trying to develop) a small python application (currently intended only for personal use) which should run on a RaspberryPi and which (one day should)

* regularly fetches the status of the heat pump
* sends messages (containing the measurements of the heat pump) to MQTT
* has a small web server with a REST API (e.g. to request changes for the heat pump)
* controls the heat pump (e.g. on a regular schedule beyond the capabilities of the heat pump)

[...]

Issue #33, opened 2022-11-22
Add support for Type: Time


rhammen:
There are a lot of parameters that have "_zeit_" in their description.
These parameters indicate a time. The value is stored as "seconds past midnight".
So e.g. 
        0 = 00:00:00 (midnight)
  3600 = 01:00:00 (1 am)
23400 = 06:30:00 (6:30 am)
43800 = 12:10:00 
72000 = 20:00:00 (8 pm)
Would it be possible to add support for this?

Bouni:
Can you give a concrete parameter that has this format so that I can check it at my pump?

rhammen:
This is about paramater with "zeit" in their description.
So for example all parameter with numbers 162-221, 223-282, 284-343, 345-404, 406-505, 507-606, 608-667.
It concerns the times you block e.g. heating or domestic water.
Example:
On your Luxtronik go to Domestic Water --> Clockprogram (not sure of the exact name, since mine is in Dutch)
Then choose 5+2, so weekdays and weekend.
You get to a screen showing "Monday-Friday" and 5 rows, each rows showing a START and END 
My first line reads: 00:00 - 12:10
These values of this first line end up in parameter 416-417. 
So my parameter 416 =0 (00:00) and parameter 417 = 43800 (12:10 = 12*3600+10*60)

If you scroll down the screen past row 5, you get to the screen "Saturday-Sunday".
My first line here reads 00:00 - 14:00.
These values of this first line end up in parameter 426-427.
So parameter 427 = 50400 (14:00 = 14*3600)
 
![20221124_124556](https://user-images.githubusercontent.com/75572839/203777096-7b9dfbfb-09f3-4bbb-8087-6a59115f11eb.jpg)
![20221124_124605](https://user-images.githubusercontent.com/75572839/203777111-b343f6b3-f8dd-4437-8277-b40773f17d93.jpg)

kbabioch:
I'm happy to work on this, although I'm rather unsure about how to best handle this within the code.

A parameter value is always 8 bytes long and cannot hold all the information that is mentioned above. That's the reason why it is spread across multiple parameter entries. This is fine in general, but so far the code and datatypes work with the assumption that conversions and representation can be done based on a single parameter.

Any ideas on how to best adjust the code here @Bouni and/or @benpru? Essentially we need a data type / structure that spreads across multiple parameters. Whenever something is changed, potentially multiple parameters need to be reset.

gerw:
[...]

PR #123, opened 2023-07-16
Create a VersionChild datatype

gerw:
This works towards #122.

In `calculations.py` we have an exceptional handling for the fields 81-90, since these ten fields contain the version information. Positions 82-90 were uninhabited; with this patch, a new `VersionChild` sits there and gobbles the value for its parent.

A new flag `datatype_internal` indicates, that this datatype is only internal and the iterator of `Calculations` will skip this entry.

### DIFF
--- a/luxtronik/calculations.py
+++ b/luxtronik/calculations.py
@@ -5,6 +5,7 @@ from luxtronik.datatypes import (
     BivalenceLevel,
     Bool,
     Celsius,
+    Character,
     Count,
     Energy,
     Errorcode,
@@ -28,7 +29,6 @@ from luxtronik.datatypes import (
     SwitchoffFile,
     Timestamp,
     Unknown,
-    Version,
     MajorMinorVersion,
     Voltage,
 )
@@ -122,7 +122,16 @@ class Calculations:
             78: HeatpumpCode("ID_WEB_Code_WP_akt"),
             79: BivalenceLevel("ID_WEB_BIV_Stufe_akt"),
             80: OperationMode("ID_WEB_WP_BZ_akt"),
-            81: Version("ID_WEB_SoftStand"),
+            81: Character("ID_WEB_SoftStand_0"),
+            82: Character("ID_WEB_SoftStand_1"),
+            83: Character("ID_WEB_SoftStand_2"),
+            84: Character("ID_WEB_SoftStand_3"),
+            85: Character("ID_WEB_SoftStand_4"),
+            86: Character("ID_WEB_SoftStand_5"),
+            87: Character("ID_WEB_SoftStand_6"),
+            88: Character("ID_WEB_SoftStand_7"),
+            89: Character("ID_WEB_SoftStand_8"),
+            90: Character("ID_WEB_SoftStand_9"),
             91: IPv4Address("ID_WEB_AdresseIP_akt"),
             92: IPv4Address("ID_WEB_SubNetMask_akt"),
             93: IPv4Address("ID_WEB_Add_Broadcast"),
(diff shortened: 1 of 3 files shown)


luxtronik/datatypes.py
```diff
@@ -333,10 +334,45 @@ class Version(Base):
...
+        """Converts value into heatpump units."""
+        raise NotImplementedError(
+            '"to_heatpump()" is not implemented for class "Version"'
+        )
+
     @classmethod
     def from_heatpump(cls, value):
         return "".join([chr(c) for c in value]).strip("\x00")
 
+    @Base.value.getter
```
gerw, line 353: `pylint` complains here. Is this the correct way to overwrite the getter of `value` in the `Base` class?

luxtronik/calculations.py
```diff
@@ -122,7 +123,16 @@ def __init__(self):
...
+            81: version,
+            82: version.spawn_child(),
+            83: version.spawn_child(),
+            84: version.spawn_child(),
+            85: version.spawn_child(),
+            86: version.spawn_child(),
+            87: version.spawn_child(),
+            88: version.spawn_child(),
+            89: version.spawn_child(),
+            90: version.spawn_child(),
```
kbabioch, line 135: - Does this datatype no longer have a name (`ID_WEB_SoftStand`)?
- I totally get that the handling of this datatype was not particularly nice and iterating over it was generating confusing results. While this fixes it, the definition doesn't look super handy. Couldn't we address this in the iterator itself somehow by indicating the length of each datatype, etc.?

gerw: * The name is unchanged, it is defined in Line 43.
* The version datatype is the only one with a length bigger than 1. Thus, it was more reasonable to me to handle it as an exceptional case by introducing the "sub"-datatype.
* To which "iterator" are you referring to?

kbabioch: > The name is unchanged, it is defined in Line 43.

Ah, somehow missed that on first glance, sorry for that.

> The version datatype is the only one with a length bigger than 1. Thus, it was more reasonable to me to handle it as an exceptional case by introducing the "sub"-datatype.

Depending on how we're going to tackle #33, we might have other datatypes with length > 1, also.

Personally I would prefer some kind of length attribute (with the default being 1), compared to this approach.

> To which "iterator" are you referring to?

In the description of this pull request you were referring to the iterator. I suppose you mean the `__iter__` method of `Calculations`. Haven't checked this in close detail yet, probably it didn't need to skip anything before, since the values have not been considered for iteration in the first place.

kbabioch:
Other input from @Bouni et al. might also be welcome ;-).

luxtronik/calculations.py
gerw: Concerning #33: In my opinion, the data there has length one (each time sits in one slot). The overall handling of the data there should be (in my opinion) handled in a higher-level way in the sense of #119.

gerw:
> There seem to be two issues with pylint (as already noted).
> 
> ```
>  luxtronik/datatypes.py:354:4: W0236: Method 'value' was expected to be 'property', found it instead as 'method' (invalid-overridden-method)
> ************* Module tests.test_datatypes
> tests/test_datatypes.py:713:15: W0143: Comparing against a callable, did you omit the parenthesis? (comparison-with-callable)
> ```

Yes, but I do not understand how to fix it.

gerw:
I realized another possibility which seems to (almost) remove the exceptional handling:

* We introduce a new datatype `character` which just converts 1 raw value in the corresponding character.
* Parameters 81-90 get this new datatype (and they will store one character of the version string)
* The version string is provided by the new higher-level interface from #119

Bouni:
@gerw That seems like the right way to solve this to me 👍🏽

kbabioch:
> I realized another possibility which seems to (almost) remove the exceptional handling:
> 
> * We introduce a new datatype `character` which just converts 1 raw value in the corresponding character.
> * Parameters 81-90 get this new datatype (and they will store one character of the version string)
> * The version string is provided by the new higher-level interface from [Provide high level interface #119](https://github.com/Bouni/python-luxtronik/issues/119)

That sounds also more "keep it simple"-like to me and introduces less complexity in the interpretation of the raw data by the current datatypes.

gerw:
@Bouni @kbabioch Done.

PR #127, opened 2023-07-20
Attempt to run CI only once

Bouni:
This is an attempt to prevent our CI pipeline from running twice for every PR.
There is no need to run it `on: push` for a PR and `on: pull-request` as well.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -2,6 +2,8 @@
 name: Lint source files
 on: # yamllint disable-line rule:truthy
   push:
+    branches:
+      - main
   pull_request:
 jobs:
   markdownlint:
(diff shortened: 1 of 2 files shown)


Bouni:
Seems to work, we run only 6 checks instead of 12 as we do currently

gerw (approved):

kbabioch:
I guess the problem with omitting `push` is that it will only run on pull requests then. This could make it harder to prepare something in a (private) branch, since the pipeline will run only on creating a pull request.

Also, what will happen when pushing additional commits to an existing pull request?

Bouni:
I guess the first point is what pre-commit  solves.

I have checked the second because I had the same concern but every push to a PR triggers the CI. ( This one has two commits and the CI ran for both)

PR #131, opened 2023-07-21
Added a decorator to connect to the heat pump

### LINKED ISSUE #130: Get rid of _read_after_write()
The connection to the heat pump is currently being established within the `_read_after_write()` function.

This makes it difficult to offer additional functions such as `read_parameters()`.

Guzz-T:
Using a decorator we can simply re-use the code to connect to the heat pump. This allows us to offers several read / write functions like only read the paramters.

Fixes #130

### DIFF
--- a/luxtronik/__init__.py
+++ b/luxtronik/__init__.py
@@ -100,7 +100,6 @@ class Luxtronik:
         with self._lock:
             self._connect()
             ret_val = func(*args, **kwargs)
-            # self._disconnect()
             return ret_val
 
     def read(self):
--- a/luxtronik/__init__.py
+++ b/luxtronik/__init__.py
@@ -63,9 +63,23 @@ class Luxtronik:
         self._port = port
         self._safe = safe
         self._socket = None
-        self.read()
+        self._connect()
 
     def __del__(self):
+        self._disconnect()
+
+    def _connect(self):
+        """Connect the socket if not already done."""
+        is_none = self._socket is None
+        if is_none:
+            self._socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        if is_none or is_socket_closed(self._socket):
+            self._socket.connect((self._host, self._port))
+            LOGGER.info(
+                "Connected to Luxtronik heat pump %s:%s", self._host, self._port
+            )
+
+    def _disconnect(self):
         if self._socket is not None:
             if not is_socket_closed(self._socket):
                 self._socket.close()
@@ -74,43 +88,50 @@ class Luxtronik:
                 "Disconnected from Luxtronik heatpump %s:%s", self._host, self._port
             )
 
-    def read(self):
-        """Read data from heatpump."""
-        return self._read_after_write(parameters=None)
-
-    def write(self, parameters):
-        """Write parameter to heatpump."""
-        return self._read_after_write(parameters=parameters)
-
-    def _read_after_write(self, parameters):
+    def _with_lock_and_connect(self, func, *args, **kwargs):
         """
-        Read and/or write value from and/or to heatpump.
-        This method is essentially a wrapper for the _read() and _write()
-        methods.
+        Decorator around various read/write functions to connect first.
+
+        This method is essentially a wrapper for the _read() and _write() methods.
         Locking is being used to ensure that only a single socket operation is
         performed at any point in time. This helps to avoid issues with the
         Luxtronik controller, which seems unstable otherwise.
-        If write is true, all parameters will be written to the heat pump
-        prior to reading back in all data from the heat pump. If write is
-        false, no data will be written, but all available data will be read
-        from the heat pump.
+        """
+        with self._lock:
+            self._connect()
+            ret_val = func(*args, **kwargs)
+            # self._disconnect()
+            return ret_val
+
+    def read(self):
+        """
+        Read data from heat pump.
+        All available data will be read from the heat pump.
+        """
+        return self._with_lock_and_connect(self._read)
+
+    def read_parameters(self):
+        """Read parameters from heat pump."""
+        return self._with_lock_and_connect(self._read_parameters)
+
+    def read_calculations(self):
+        """Read calculations from heat pump."""
+        return self._with_lock_and_connect(self._read_calculations)
+
+    def read_visibilities(self):
+        """Read visibilities from heat pump."""
+        return self._with_lock_and_connect(self._read_visibilities)
+
+    def write(self, parameters):
+        """
+        Write parameter to heat pump.
+        All parameters will be written to the heat pump
+        prior to reading back in all data from the heat pump.
         :param Parameters() parameters  Parameter dictionary to be written
                           to the heatpump before reading all available data
-                          from the heatpump. At 'None' it is read only.
+                          from the heat pump.
         """
-
-        with self._lock:
-            is_none = self._socket is None
-            if is_none:
-                self._socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
-            if is_none or is_socket_closed(self._socket):
-                self._socket.connect((self._host, self._port))
-                LOGGER.info(
-                    "Connected to Luxtronik heatpump %s:%s", self._host, self._port
-                )
-            if parameters is not None:
-                return self._write(parameters)
-            return self._read()
+        return self._with_lock_and_connect(self._write, parameters)
 
     def _read(self):
         parameters = self._read_parameters()


gerw:
Shouldn't it be possible to use your `_connect` as a "real" decorator (with `@`) around the read/write methods?

Guzz-T:
Unfortunately, a "real" decorator cannot use `self` as an argument. Without `self`, however, the internal fields cannot be accessed correctly. Also `pylint` complains.

gerw:
I tried a little bit and it seems to be possible with a "real" decorator:
````python
def decorator(func):
    def _decorator(self, *args, **kwargs):
        self.num_calls += 1
        return func(self, *args, **kwargs)
    return _decorator

class TestSample:
    def __init__(self):
        self.num_calls = 0

    @decorator
    def dummy(self, foo):
        print(foo)

t = TestSample()

print(f"t.num_calls: {t.num_calls}")
t.dummy(2)
print(f"t.num_calls: {t.num_calls}")
````
There are also no complaints by pylint (only concerning missing docstrings and so on).

**Comment by Bouni** [comment]
@gerw please push a CI trigger commit to check if #134 solved the pytest issue

**Comment by Guzz-T** [comment]
Have you tried accessing private fields like `self._num_calls` instead of `self.num_calls`?

**Comment by gerw** [comment]
Yes, the following is `pylint`-compatible:
```python
"""Random module"""
def decorator(func):
    """Random decorator"""
    def _decorator(self, *args, **kwargs):
        self._num_calls += 1
        return func(self, *args, **kwargs)
    return _decorator

class TestSample:
    """Random class"""
    def __init__(self):
        self._num_calls = 0

    def another_method(self):
        """Random member fctn"""
        return self._num_calls

    @decorator
    def dummy(self, foobar):
        """Random member fctn"""
        print(foobar)

t = TestSample()

print(t.another_method())
t.dummy(2)
print(t.another_method())
```

luxtronik/__init__.py
```diff
@@ -74,43 +74,60 @@ def __del__(self):
...
+                        "Connected to Luxtronik heatpump %s:%s", self._host, self._port
+                    )
+                return func(*args, **kwargs)
+
+        return inner
+
+    def read(self):
+        """
+        Read data from heatpump.
+        All available data will be read from the heat pump.
```
kbabioch, line 105: We should agree on a spelling here, i.e. `heat pump` vs. `heatpump`. Seems like `heat pump` is the "more correct" spelling.

```suggestion
        Read data from heat pump.
        All available data will be read from the heat pump.
```

luxtronik/__init__.py
```diff
@@ -74,43 +74,60 @@ def __del__(self):
...
+
+        def inner(*args, **kwargs):
+            with self._lock:
+                is_none = self._socket is None
+                if is_none:
+                    self._socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+                if is_none or is_socket_closed(self._socket):
+                    self._socket.connect((self._host, self._port))
+                    LOGGER.info(
+                        "Connected to Luxtronik heatpump %s:%s", self._host, self._port
```
kbabioch, line 96: We should agree on a spelling here, i.e. `heat pump` vs. `heatpump`. Seems like `heat pump` is the "more correct" spelling.

```suggestion
                        "Connected to Luxtronik heat pump %s:%s", self._host, self._port
```

luxtronik/__init__.py
```diff
@@ -74,43 +74,60 @@ def __del__(self):
...
+
+    def read(self):
+        """
+        Read data from heatpump.
+        All available data will be read from the heat pump.
+        """
+        return self._connect(self._read)()
+
+    def read_parameters(self):
+        """Read parameters from heatpump."""
```
kbabioch, line 110:
```suggestion
        """Read parameters from heat pump."""
```

luxtronik/__init__.py
```diff
@@ -74,43 +74,60 @@ def __del__(self):
...
+        return self._connect(self._read_parameters)()
+
+    def read_calculations(self):
+        """Read calculations from heatpump."""
```
kbabioch, line 114:
```suggestion
        """Read calculations from heat pump."""
```

luxtronik/__init__.py
```diff
@@ -74,43 +74,60 @@ def __del__(self):
...
+        return self._connect(self._read_calculations)()
+
+    def read_visibilities(self):
+        """Read visibilities from heatpump."""
```
kbabioch, line 118:
```suggestion
        """Read visibilities from heat pump."""
```

luxtronik/__init__.py
```diff
@@ -74,43 +74,60 @@ def __del__(self):
...
+        return self._connect(self._read_visibilities)()
+
+    def write(self, parameters):
+        """
+        Write parameter to heatpump.
```
kbabioch, line 123:
```suggestion
        Write parameter to heat pump.
```

luxtronik/__init__.py
```diff
@@ -74,43 +74,60 @@ def __del__(self):
...
+        All parameters will be written to the heat pump
+        prior to reading back in all data from the heat pump.
         :param Parameters() parameters  Parameter dictionary to be written
                           to the heatpump before reading all available data
-                          from the heatpump. At 'None' it is read only.
+                          from the heatpump.
```
kbabioch, line 128:
```suggestion
                          from the heat pump.
```

kbabioch:
@Guzz-T Could you elaborate on what the vision here is? How do you envision to use this in the future?

Guzz-T:
> 

Does the comment in #130 answer your question?

Guzz-T:
> @Guzz-T Could you elaborate on what the vision here is? How do you envision to use this in the future?

Does the comment in #130 answer your question?

Guzz-T:
> Yes, the following is `pylint`-compatible:

@gerw: I followed your suggestions, but now we get several `W0212 `pylint warnings

gerw:
This is strange. Pylist does not complain about something like `self._num_calls += 1`, but about `self._num_calls = self._num_calls + 1`.

Guzz-T:
> This is strange. Pylist does not complain about something like `self._num_calls += 1`, but about `self._num_calls = self._num_calls + 1`.

I reverted the changes because it would just be a different style of the decorator

gerw:
Since we are not saving the decorated function, it seems not necessary to implement this as a decorator. What about something like
```python
    def _with_lock_and_connect(self, func, *args, **kwargs):
        with self._lock:
            is_none = self._socket is None
            if is_none:
                self._socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
            if is_none or is_socket_closed(self._socket):
                self._socket.connect((self._host, self._port))
                LOGGER.info(
                    "Connected to Luxtronik heat pump %s:%s", self._host, self._port
                )
            return func(*args, **kwargs)


    def read(self):
        return self._with_lock_and_connect(self._read)
```
?

Guzz-T:
There isn't just one way of doing it. My version works like a "@"-decorator.

luxtronik/__init__.py
```diff
@@ -74,43 +88,50 @@ def __del__(self):
...
         Luxtronik controller, which seems unstable otherwise.
-        If write is true, all parameters will be written to the heat pump
-        prior to reading back in all data from the heat pump. If write is
-        false, no data will be written, but all available data will be read
-        from the heat pump.
+        """
+        with self._lock:
+            self._connect()
+            ret_val = func(*args, **kwargs)
+            # self._disconnect()
```
gerw, line 103: This line can be deleted?

luxtronik/__init__.py
```diff
@@ -63,9 +63,23 @@ def __init__(self, host, port=LUXTRONIK_DEFAULT_PORT, safe=True):
         self._port = port
         self._safe = safe
         self._socket = None
-        self.read()
+        self._connect()
 
     def __del__(self):
+        self._disconnect()
+
+    def _connect(self):
```
gerw, line 71: I am not sure, but should 'self._lock' also used in the `_connect` method?

gerw: (and the same for `_disconnect`?)

luxtronik/__init__.py
Guzz-T: Done. Btw.: It might be better to close the connection afterwards

luxtronik/__init__.py
Guzz-T: `_connect` is called inside `_lock_and_connect` for for multithreaded access to the instance. `__init__` and `__del__` are only called by one thread.

gerw (approved):

PR #134, opened 2023-07-22
Update pytest.yml

gerw:
According to https://github.com/actions/first-interaction/issues/10#issuecomment-670968624, this should fix the issue with the failing of pytest on pull requests from forks.


gerw:
(In this PR, pytest succeeded, since the branch comes from this repo and not from a fork)

Bouni (approved):
Seems right, let's test it

PR #137, opened 2023-09-19
Update parameters.py

02curls:
edited the 14,15,16 and 774,775,776 triplets, these are independent internal heating circuits parameters analogous with 11,12,13 triplet,  11 is analogous to 14 and 774
12 is analogous to 15 and 775
13 is analogous to 16 and 776


gerw:
Which part of the heating system is actually controlled by these parameters? I guess that `MK1` stands for `Mischkreis 1`, but what does it exactly mean? There are also the parameters 141-143 which again should have a similar meaning.

02curls:
@gerw  

the triplet 11,12,13 defines the heating curve of the core internal heating circuit (and the accumulation tank in my case), 
the next analogous triplet 14,15,16 defines the heating curve of internal heating circuit Mix 1 (my first floor heating circuit in my case), 
and the next one 774,775,776 defines the heating curve of internal heating circuit Mix3 (my second floor heating circuit in my case)

 Parameters 11 and analogously 14, 774 have the following min:20 max:70 step:0,5 
11name: luxtronik2_heating_circuit_curve1_temperature 
14name could be: luxtronik2_heating_circuitMix1_curve1_temperature 
774name  could be: luxtronik2_heating_circuitMix3_curve1_temperature

   Parameters 12 and analogously 15, 775 have min:5 max:35 step:0,5 
12name: luxtronik2_heating_circuit_curve2_temperature 
15name could be: luxtronik2_heating_circuitMix1_curve2_temperature 
775name could be: luxtronik2_heating_circuitMix3_curve2_temperature

  Parameters 13 and analogously 16, 776 (min:-15 max:10 step:0,5 
13name: luxtronik2_heating_circuit_curve_night_temperature 
16name could be: luxtronik2_heating_circuitMix1_curve_night_temperature 
776name could be: luxtronik2_heating_circuitMix3_curve_night_temperature

I guess 141-143 will be analogous

printscreen from my https://www.heatpump24.com/Einstellung_102.php?WPID=59………

Screenshot 2023-09-20 at 6 13 02

kbabioch:
> Which part of the heating system is actually controlled by these parameters? I guess that `MK1` stands for `Mischkreis 1`, but what does it exactly mean?

AFAIK this will be different in different installations. Typically there are different heating systems, so that you need to mix your water to different temperatures ("mischen").

In my case, for instance, `MK1` is used for the cooling functionality.

> the triplet 11,12,13 defines the heating curve of the core internal heating circuit (and the accumulation tank in my case),
the next analogous triplet 14,15,16 defines the heating curve of internal heating circuit Mix 1 (my first floor heating circuit in my case),
and the next one 774,775,776 defines the heating curve of internal heating circuit Mix3 (my second floor heating circuit in my case)

What do you mean by heating curve? What exactly do these curve describe?

Kars-de-Jong:
@gerw The function of the mixing circuits depends on the specific heat pump and the options installed at the customer. So there is no fixed rule as to what they are used for. Their use is specified with parameters such as `ID_Einst_MK2Typ_akt`. which is an enum. Not every mixing circuit has the same options. The possible valies for mixing circuit 2 for instance are (taken from the web socket interface, in Dutch, sorry about that):

```xml
<item id="0x42434cd4">
    <name>Menggroep 2</name>
    <option value="0">Nee</option>
    <option value="1">Ontladen</option>
    <option value="2">Laden</option>
    <option value="3">Koelen</option>
    <option value="4">Verw.+koel</option>
    <raw>3</raw>
    <value>Koelen</value>
</item>
```
As read using the library:
```
Number: 130   Name: ID_Einst_MK2Typ_akt                                          Type: Unknown              Value: 3
```

Kars-de-Jong:
> What do you mean by heating curve? What exactly do these curve describe?

The heating curve determines the target temperature in the mixing circuit based on another value, usually the outside temperature. The parameters determine the start and the end of the curve.

Kars-de-Jong:
Possible options for `ID_Einst_MK1Typ_akt`:
```xml
<item id="0x4242a66c">
    <name>Menggroep 1</name>
    <option value="0">Nee</option>
    <option value="1">Ontladen</option>
    <option value="2">Laden</option>
    <option value="3">Koelen</option>
    <option value="4">Verw.+koel</option>
    <raw>0</raw>
    <value>Nee</value>
</item>
```

(it is not used in my system).

And possible options for `ID_Einst_MK3Typ_akt`:
```xml
<item id="0x42439534">
    <name>Menggroep 3</name>
    <option value="0">Nee</option>
    <option value="1">Ontladen</option>
    <option value="4">Verw.+koel</option>
    <raw>0</raw>
    <value>Nee</value>
</item>
```

(it is also not used in my system).

02curls:
@kbabioch  heating curve on an alpha innotec: https://www.youtube.com/watch?v=7WS2MhA43ns

BenPru (approved):
LGTM

gerw (approved):

kbabioch:
Changes look good to me, too.

However I'm wondering:

Shouldn't this be applied to `Mk2` also:

https://github.com/Bouni/python-luxtronik/blob/9470de3267fad8c5f9aa3cb379b37f292044fcca/luxtronik/parameters.py#L182-L184

02curls:
@kbabioch I believe you are right, the same changes should be applied to MK2 also,
since I am new to GitHub, I am not sure, how to resubmit the parameters.py with MK2 changes applied (here: [https://www.dropbox.com/scl/fi/o5xz3nhzj0fyvtne5v4w8/parameters.py?rlkey=9mtqm5tvd59u5rovre21gb983&dl=0](url)

PR #141, opened 2023-10-06
More parameters identified as energies

gerw:
Three more parameters are identified as energies.


luxtronik/parameters.py
```diff
@@ -893,9 +893,9 @@ def __init__(self, safe=True):
             849: Unknown("ID_Ba_Hz_MK3_saved"),
             850: Hours("ID_Einst_Kuhl_Zeit_Ein_akt", True),
             851: Hours("ID_Einst_Kuhl_Zeit_Aus_akt", True),
-            852: Unknown("ID_Waermemenge_Seit"),
+            852: Energy("ID_Waermemenge_Seit"),
```
kbabioch, line 896: Do we know what `Seit` refers to here? Are we sure this is an Energy and not some Timestamp value?

luxtronik/parameters.py
```diff
@@ -893,9 +893,9 @@ def __init__(self, safe=True):
...
             853: Unknown("ID_Waermemenge_WQ"),
-            854: Unknown("ID_Waermemenge_Hz"),
+            854: Energy("ID_Waermemenge_Hz"),
```
kbabioch, line 898: `Hz` probably means Heizung in this context?

luxtronik/parameters.py
```diff
@@ -1101,7 +1101,7 @@ def __init__(self, safe=True):
             1057: Unknown("ID_Einst_Vorlauf_ZUP"),
             1058: Unknown("ID_Einst_Abtauen_im_Warmwasser"),
             1059: Energy("ID_Waermemenge_ZWE"),
-            1060: Unknown("ID_Waermemenge_Reset"),
+            1060: Energy("ID_Waermemenge_Reset"),
```
kbabioch, line 1104: This is probably the heat amount since the last reset?

luxtronik/parameters.py
gerw: Yes. This seems to be the energy since the time stamp from parameter 880.

luxtronik/parameters.py
gerw: Yes, I think so.

luxtronik/parameters.py
gerw: Yes, I think a reset can be performed via the menu of the heat pump. (So I do not think that this is a reset of the heat pump itself.)

kbabioch: Yes, I meant the reset via the menu.

kbabioch (approved):

PR #142, opened 2023-10-09
Use round instead of int in `datatypes.py`

gerw:
`int(some_float)` always rounds towards zero and this is (maybe) not what is expected.

This PR suggests to replace `int()` by `round()`.


Bouni (approved):
Lgtm

PR #145, opened 2024-02-03
Add some heatpump codes

gerw:


gerw:
Does anybody know, why `black` is complaining? Has the configuration or the `black´ version changed?

Bouni:
I guess its a version mismatch between the pre-commit-config and the CI

CI:

```
Run pip install black
Collecting black
  Downloading black-24.1.1-cp311-cp311-manylinux_2_17_x86_64.manylinux2014_x86_64.whl.metadata (73 kB)
```

pre-commit
```
repos:
  - repo: https://github.com/psf/black
    rev: 22.12.0
```

luxtronik/datatypes.py
```diff
@@ -472,6 +472,14 @@ class HeatpumpCode(SelectionBase):
...
         30: "WZWD",
         31: "ERC",
+        32: "ERC",
+        33: "ERC",
+        34: "ERC",
+        35: "ERC",
+        36: "ERC",
+        37: "ERC",
+        38: "ERC",
+        39: "ERC",
```
Bouni, line 482: Is this really 8 time the same code?

gerw: Yes, this is what you would see in the luxtronik controller for a heat pump of type 31-39. I am not sure, if one can find any of these in the wild...

Bouni: Does not sound unfamiliar for AIT to do such a mess 😅 I was just curious if it was kind of a typo/copy and paste error

Bouni:
@gerw If you could review #149 and rebase after the merge I will merge this one

luxtronik/datatypes.py
kbabioch: What is `ERC`? Does anyone know?

luxtronik/datatypes.py
```diff
@@ -504,10 +512,23 @@ class HeatpumpCode(SelectionBase):
         69: "MSW 6S",
         70: "MSW 8S",
         71: "MSW 10S",
-        72: "MSW 13S",
+        72: "MSW 12S",
```
kbabioch, line 515: Are we sure about this change? Where was this originally coming from?

gerw: The `13S` seems to be present in all open source projects reading data from the luxtronik controller. But the luxtronik controller itself would present these heat pumps as `12S`.  Google doesn't find anything related to these strings...

luxtronik/datatypes.py
gerw: No idea. Maybe this is just some leftover...

Bouni:
@gerw Can you please rebase this one as well?

gerw:
@Bouni: Hurray! :)

Do you know why the bot did not edit the old commit? Is it due to the fact that the old commit was spawned from another work flow? Maybe I just push another time to this branch to see what is happening.

gerw:
Yes, this seems to be the case. The second coverage comment has been edited after the latest push. :partying_face:

luxtronik/datatypes.py
Bouni: I've looked around and every source I found says something like ERC = unknown. I would merge this and if some day in the future the real meaning comes to light we can fix that.

luxtronik/datatypes.py
Bouni: I did a quick check against the `appl` binary of the heatpump and there I find only `12S` but n `13S`:

```sh
strings firmware/appl | grep 12S$
MSW_3_12S
MSW12S
MSW3-12S
MSW_3_12S
```

```sh
strings firmware/appl | grep 13S$
```

So I guess `12S` is right

Bouni (approved):

PR #152, opened 2024-02-05
Try traget pull_request instead of pull_request_target

Bouni:
This is an attempt to fix the pytest issue surfaced in #151

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -4,13 +4,15 @@ on: # yamllint disable-line rule:truthy
   push:
     branches:
       - main
-  pull_request_target:
+  pull_request:
 jobs:
   pytest:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout code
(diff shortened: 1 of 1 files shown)


Bouni:
@gerw My guess was wrong, this prints `(HEAD detached at pull/152/merge) 1fcdeb4 Merge c97a155cc37e44e42515b7721cf4b8340826235d into 3d94a5903fe1e104531bd481e4277dfc11668f1a` for `git branch -v` 😕 

@kbabioch Do you have an idea why the test seem to run on the wrong files ?

Bouni:
Ok I think `pull_request_target` is indeed the problem, the tests for the trigger `pull_request` succeed!

Bouni:
I would like to merge the pytest CI workflow with `pull_request_target` replaced with `pull_request` and see if that works for #151 as well.

Any thoughts?

gerw:
LGTM. Thank you!

gerw (approved):

PR #154, opened 2024-02-06
Fix coverage

Bouni:
Next attempt, this time I try what Chat GPT thinks is the right way of doing this


Bouni:
@gerw Not sure if this does the trick. Would you mind doing another merge and rebase?

gerw (approved):
Let's try.

gerw:
Nope. From my very little understanding, it seems that the git workflow is running in my repository (for a PR from my fork), and this does not have access to stuff from this repo (e.g., to `GITHUB_TOKEN`). But I might be wrong.

PR #157, opened 2024-02-07
Fix workflow_run name

Bouni:
`workflows:`must contain the `name` of the other workflow, not the filename


Bouni:
@gerw you can merge this if you have time to do so

gerw (approved):

PR #159, opened 2024-02-08
Fix coverage comment

Bouni:
I finally fixed the coverage comment issue 🥳 

1. It is necessary to have a two staged solution because of security reasons, details can be found in GitHubs blog post: https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

2. MishaKav/pytest-coverage-comment did not allow the `workflow_run` trigger which prevented the coverage report to be posted. I forked the repo, fixed the issue and opend https://github.com/MishaKav/pytest-coverage-comment/pull/153 . Until that PR is merged we can use my fork `bouni/pytest-coverage-comment@workflow_run`

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -1,5 +1,5 @@
 ---
-name: Comment coverage report on the pull request
+name: PR coverage comment
 on: # yamllint disable-line rule:truthy
   workflow_run:
     workflows: ["Run unit tests"]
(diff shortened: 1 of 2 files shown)


gerw:
If I understand correctly, the `pytest` workflow also runs after each push to the `main` branch. Which number is then in the artifact `pr/PR-number.txt`? What happens in the `coverage` workflow afterwards?

I think that we need two workflows:
* on push, only the tests need to run
* on PR, the coverage report should be published as a comment

Maybe this can be implemented in one `pytest` workflow by some branching?

Bouni:
@gerw I looked into it and it actually doen't matter what ends up in the PR-number.txt because in the coverage.yml we check if the trigger was a pull_request: https://github.com/Bouni/python-luxtronik/blob/738ba1f4d1507e006dd514f78914b81158c9f7e9/.github/workflows/coverage.yml#L12-L14

The workflow does not run on other events.

gerw (approved):

Issue #158, opened 2024-02-07
Are frequent parameter changes dangerous?


gerw:
Recently, I came across an interesting [post](https://forum.fhem.de/index.php?msg=863229) which reads

> I can't comment much on older Luxtronik versions, but if you would reverse engineer, you would see that the 2.1 Luxtronik has a Micron 1.8V 128MB NAND, which pretty certainly means it's one of those 100k erase count chips, and trust me they do balance the eraseblocks. There is an Atmel ARM-board in the luxtronik. Without going too much into the details my take is that changing a setting once or twice a day isn't going to wear that thing out any time soon, but I wouldn't go excessive on those settings changes.

Similarly, a [Finnish post](https://www.maalampofoorumi.fi/index.php?topic=7639.285;wap2) (maybe by the same person) translates to

>The little birds sang that Luxtronik has Micron's 128 megabyte 1.8V Nand flash, which based on that voltage is most likely SLC technology and lasts up to 100,000 erase cycles per erase block. I hear that on average there are not fifty writes per erase block per year, so we can get to the point that the luxtronik will never die in terms of the number of writes. Even if you fiddle with its settings several times a day, it won't.
>
>So I think tuning the relays is pointless. That Nand is very durable, something the Germans have done right.

Now I am wondering what "I wouldn't go excessive" really means.

I tried to make up some numbers. The first thing I don't know is the size of the erase blocks. In what follows, I use 128kb, but I might be totally wrong. Maybe we can use the ~~feature~~bug and get this information via ssh? If we have 100k erase cycles per erase block, this means that we have 100M "write actions" (each one writing one erase block).

The next thing that we need to know is how many periodically occurring write actions we have.
* Every two hours, the controller writes the DTA file `NewProc` (size 350 kb). [This can be seen by looking on the time stamp, since this file is served by the web server.] This means 3 write actions + file system overhead.
* For each parameter change, the file `appl_param2` is removed, `appl_param1` is moved to `appl_param2` and `appl_param1` is rewritten. [Again, these files are served by the the server.] It is not clear to me, how many write actions this needs (in particular, the file system overhead again).
* Every two hours, there is a parameter update (related to `Heizgrenze`), maybe only in the heating period?

Is there any idea how to estimate the file system overhead? Would it help to know the file system?

I think there are no other frequently occurring write actions. I think we can safely ignore firmware updates and other stuff which only happens now and then.

[...]

PR #162, opened 2024-02-12
Update README.md

gerw:
Addresses #158.

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,9 +210,12 @@ from luxtronik import Luxtronik
 l = Luxtronik('192.168.1.23', 8889, safe=False)
 ```
 
-**CAUTION:** The heat pump controller uses a NAND memory chip which has a limited number of erase cycles. 
-Since every change of the parameter induces some file changes on the controller, frequent parameter changes
-may limit the lifetime of your heat pump, see the discussions in [issue 158](https://github.com/Bouni/python-luxtronik/issues/158).
+**CAUTION:** The heat pump controller uses a NAND memory chip which has a
+limited number of erase cycles.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -210,6 +210,13 @@ from luxtronik import Luxtronik
 l = Luxtronik('192.168.1.23', 8889, safe=False)
 ```
 
+**CAUTION:** The heat pump controller uses a NAND memory chip which has a
+limited number of erase cycles.
+Since every change of the parameter induces some file changes on the
+controller, frequent parameter changes
+may limit the lifetime of your heat pump, see the discussions in
+[issue 158](https://github.com/Bouni/python-luxtronik/issues/158).
+
```
kbabioch, line 219: In order to stay consistent with other NOTEs in this README, I would suggest to use `NOTE`. Also writing by itself is not an issue we want to warn about, just the amount of writes, since some changes are needed / wanted.



```suggestion
**NOTE:** The heat pump controller uses a NAND memory chip as persistent storage.
This technology has only a limited number of erase cycles.
As far as we know, every change of a parameter will eventually induce some file changes on the controller,
hence **frequent parameter changes may limit the lifetime of your heat pump**.
For further details refer to [this discussion](https://github.com/Bouni/python-luxtronik/issues/158).
```

gerw: Thank you for your suggestion. I adopted it, but removed the "as far as we know", since one can be pretty certain that file changes occur (look at the time stamps of `http://your_heat_pump/appl_data1` and `http://your_heat_pump/appl_data2`).

kbabioch (approved):

PR #163, opened 2024-02-13
Issue-132: Revert API changes introduces with #86

### LINKED ISSUE #132: Parameters (and friends) should be members of Luxtronik
One of the changes of pull request #86 was to remove `parameters`, `calculations` and `visibilities` as members of `Luxtronik`. This change was not needed to arrive at the goal of #85 to support multiple instances of `Luxtronik` (this was solved by making the parameters instance members of `Parameters`).

In order to ease the implementation of a high-level interface for reading (see #119), I suggest to add back `parameters` as an instance member of `Luxtronik`. Otherwise, I do not see an easy way for providing functions like `get_firmware_version` to `Luxtronik`.

Guzz-T:
With #86 we split the data from the read/write interface.
However, these changes mean that the existing program needs to be adjusted.

With this pull-request we restore the previous api by:
- Add a data class that contains the parameters, calculations and visibilities
- Keep the seperation of the data class and the read/write interface
- Provide a new "Luxtronik" class that combines the data and the interface

This allows both the old and the new API to be used.

Fix #132 

NOTE: The linter added some format changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,26 +53,37 @@ The following example reads in data from the heat pump:
 from luxtronik import Luxtronik
 
 l = Luxtronik('192.168.1.23', 8889)
-calculations, parameters, visibilities = l.read()
+heating_limit = l.parameters.get("ID_Einst_Heizgrenze_Temp")
 
-t_forerun = calculations.get("ID_WEB_Temperatur_TVL")
+# Do something else here...
+
+# Read the values again
+l.read()
+t_forerun = l.calculations.get("ID_WEB_Temperatur_TVL")
+t_outside = l.calculations.get("ID_WEB_Temperatur_TA")
 
 # alternatively get also works with numerical ID values
-
-t_forerun = calculations.get(10)
+t_forerun = l.calculations.get(10)
 
 print(t_forerun) # this returns the temperature value of the forerun, 22.7 for example
 print(t_forerun.unit) # gives you the unit of the value if known, °C for example
 
 # calculations holds measurement values
-# check https://github.com/Bouni/luxtronik/blob/master/luxtronik/calculations.py for values you might need
+# check https://github.com/Bouni/python-luxtronik/blob/master/luxtronik/calculations.py for values you might need
 
 # parameters holds parameter values
-# check https://github.com/Bouni/luxtronik/blob/master/luxtronik/parameters.py for values you might need
+# check https://github.com/Bouni/python-luxtronik/blob/master/luxtronik/parameters.py for values you might need
 
 # visibilitys holds visibility values, the function of visibilities is not clear at this point
-# check https://github.com/Bouni/luxtronik/blob/master/luxtronik/visibilities.py for values you might need
+# check https://github.com/Bouni/python-luxtronik/blob/master/luxtronik/visibilities.py for values you might need
 ```
+
+The method `read()` reads the calculations, parameters and
+visibilities from the heat pump.
+Alternatively `read_parameters()`, `read_calculations()` or `read_visibilities()`
+can be used.
+
+Note that an initial read operation is carried out in the constructor.
 
 ### SCRIPTS AND COMMAND LINE INTERFACE (CLI)
 
@@ -188,8 +199,16 @@ from luxtronik import Luxtronik, Parameters
 
 l = Luxtronik('192.168.1.23', 8889)
 
+# Queue a parameter change
+# In this example, the domestic hot water temperature is set to 45 degrees.
+l.parameters.set("ID_Soll_BWS_akt", 45.0)
+
+# Write all queued changes to the heat pump
+l.write()
+
+# Another possibility to write parameters
 parameters = Parameters()
-heating_mode = parameters.set("ID_Ba_Hz_akt", "Party")
+parameters.set("ID_Ba_Hz_akt", "Party")
 l.write(parameters)
 
 # If you're not sure what values to write, you can get all available options:
--- a/luxtronik/__init__.py
+++ b/luxtronik/__init__.py
@@ -38,9 +38,7 @@ def is_socket_closed(sock: socket.socket) -> bool:
     """Check is socket closed."""
     try:
         # this will try to read bytes without blocking and also without removing them from buffer
-        data = sock.recv(
-            LUXTRONIK_SOCKET_READ_SIZE_PEEK, socket.MSG_DONTWAIT | socket.MSG_PEEK
-        )
+        data = sock.recv(LUXTRONIK_SOCKET_READ_SIZE_PEEK, socket.MSG_DONTWAIT | socket.MSG_PEEK)
         if len(data) == 0:
             return True
     except BlockingIOError:
@@ -48,21 +46,30 @@ def is_socket_closed(sock: socket.socket) -> bool:
     except ConnectionResetError:  # pylint: disable=broad-except
         return True  # socket was closed for some other reason
     except Exception as err:  # pylint: disable=broad-except
-        LOGGER.exception(
-            "Unexpected exception when checking if socket is closed", exc_info=err
-        )
+        LOGGER.exception("Unexpected exception when checking if socket is closed", exc_info=err)
         return False
     return False
 
 
-class Luxtronik:
-    """Main luxtronik class."""
+class LuxtronikData:
+    """
+    Collection of parameters, calculations and visiblities.
+    Also provide some high level access functions to their data values.
+    """
 
-    def __init__(self, host, port=LUXTRONIK_DEFAULT_PORT, safe=True):
+    def __init__(self, parameters=None, calculations=None, visibilities=None, safe=True):
+        self.parameters = Parameters(safe) if parameters is None else parameters
+        self.calculations = Calculations() if calculations is None else calculations
+        self.visibilities = Visibilities() if visibilities is None else visibilities
+
+
+class LuxtronikSocketInterface:
+    """Luxtronik read/write interface via socket."""
+
+    def __init__(self, host, port=LUXTRONIK_DEFAULT_PORT):
         self._lock = threading.Lock()
         self._host = host
         self._port = port
-        self._safe = safe
         self._socket = None
         self._connect()
 
@@ -76,18 +83,15 @@ class Luxtronik:
             self._socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
         if is_none or is_socket_closed(self._socket):
             self._socket.connect((self._host, self._port))
-            LOGGER.info(
-                "Connected to Luxtronik heat pump %s:%s", self._host, self._port
-            )
+            LOGGER.info("Connected to Luxtronik heat pump %s:%s", self._host, self._port)
 
     def _disconnect(self):
+        """Disconnect the socket if not already done."""
         if self._socket is not None:
             if not is_socket_closed(self._socket):
                 self._socket.close()
             self._socket = None
-            LOGGER.info(
-                "Disconnected from Luxtronik heatpump %s:%s", self._host, self._port
-            )
+            LOGGER.info("Disconnected from Luxtronik heatpump %s:%s", self._host, self._port)
 
     def _with_lock_and_connect(self, func, *args, **kwargs):
         """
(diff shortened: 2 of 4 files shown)


luxtronik/__init__.py
```diff
@@ -103,26 +116,30 @@ def _with_lock_and_connect(self, func, *args, **kwargs):
...
+        if data is None:
+            data = LuxtronikData.new(self._safe)
+        return self._with_lock_and_connect(self._read, data)
 
-    def read_parameters(self):
+    def read_parameters(self, parameters=None):
         """Read parameters from heat pump."""
-        return self._with_lock_and_connect(self._read_parameters)
+        if parameters is None:
+            parameters = Parameters(self._safe)
```
gerw, line 131: I am not sure if `LuxtronikInterface` should hold the `_safe` flag. I mean, this is not related to the interface itself. I propose that `read_parameters` just create `Parameters()` (with the default `safe = True`) if no `parameters` are passed.

luxtronik/__init__.py
```diff
@@ -38,24 +39,40 @@ def is_socket_closed(sock: socket.socket) -> bool:
...
+    @classmethod
+    def new(cls, safe) -> LuxtronikData:
+        return LuxtronikData(Parameters(safe), Calculations(), Visibilities())
+
+    @classmethod
+    def copy(cls, data) -> LuxtronikData:
+        return LuxtronikData(copy(data.parameters), copy(data.calculations), copy(data.visibilities))
+
+
+class LuxtronikInterface:
```
gerw, line 75: `LuxtronikSocketInterface` would be more descriptive?

luxtronik/__init__.py
```diff
@@ -233,3 +252,40 @@ def _read_char(self):
...
         return struct.unpack(">b", reading)[0]
+
+
+class Luxtronik(LuxtronikData):
+    def __init__(self, host, port=LUXTRONIK_DEFAULT_PORT, safe=True):
+        super().__init__(Parameters(safe), Calculations(), Visibilities())
+        self.interface = LuxtronikInterface(host, port, safe)
+        self.read()
+
+    def read(self) -> LuxtronikData:
```
gerw, line 263: Remove typing.

luxtronik/__init__.py
```diff
@@ -233,3 +252,40 @@ def _read_char(self):
...
+        """
+        Read data from heat pump.
+        All available data will be read from the heat pump.
+        """
+        self.interface.read(self)
+        return LuxtronikData.copy(self)
```
gerw, line 269: Do we really need to `copy()` here? If somebody wants to have a copy, he can do a copy himself.

luxtronik/__init__.py
```diff
@@ -38,24 +39,40 @@ def is_socket_closed(sock: socket.socket) -> bool:
...
+    Also provide some high level access functions to their data values.
+    """
+
+    def __init__(self, parameters, calculations, visibilities):
+        self.parameters = parameters
+        self.calculations = calculations
+        self.visibilities = visibilities
+
+    @classmethod
+    def new(cls, safe) -> LuxtronikData:
```
gerw, line 67: Wouldn't it be sufficient to have default values `parameters = Parameters()` (and so on) in `__init__`?

luxtronik/__init__.py
Guzz-T: Done. Currently, there is still the parameter at `Luxtronik()`

luxtronik/__init__.py
Guzz-T: Done

luxtronik/__init__.py
Guzz-T: Done

luxtronik/__init__.py
Guzz-T: Python allows only one `__init__`. In order to offer different constructors like `copy()` and `new()` we need a common implemenation for `__init__`.

Guzz-T:
Unfortunately `read_parameters(self, parameters=Parameters(True)` doesn't work reliably. The first time an instance is created. From the second call onwards, the reference is still stored, which means that no new instance is created.

gerw:
But this is fixed by moving `Parameters(True)` to the function body in the latest patch, right?

Btw: `True` is the default argument of `Parameters()` and can be omitted.

The reason seems to be the following, I didn't know that either.
> Python’s default arguments are evaluated once when the function is defined, not each time the function is called (like it is in say, Ruby). This means that if you use a mutable default argument and mutate it, you will and have mutated that object for all future calls to the function as well.

[Reference](https://docs.python-guide.org/writing/gotchas/)

luxtronik/__init__.py
```diff
@@ -38,31 +39,46 @@ def is_socket_closed(sock: socket.socket) -> bool:
...
+        self.calculations = calculations
+        self.visibilities = visibilities
+
+    @classmethod
+    def new(cls, safe=True):
+        return LuxtronikData(Parameters(safe), Calculations(), Visibilities())
+
+    @classmethod
+    def copy(cls, data):
+        return LuxtronikData(copy(data.parameters), copy(data.calculations), copy(data.visibilities))
```
gerw, line 72: The `copy` method is no longer needed?

luxtronik/__init__.py
```diff
@@ -38,31 +39,46 @@ def is_socket_closed(sock: socket.socket) -> bool:
...
-class Luxtronik:
-    """Main luxtronik class."""
+class LuxtronikData:
+    """
+    Collection of parameters, calculations and visiblities.
+    Also provide some high level access functions to their data values.
+    """
 
-    def __init__(self, host, port=LUXTRONIK_DEFAULT_PORT, safe=True):
+    def __init__(self, parameters, calculations, visibilities):
```
gerw, line 61: Can't you add default parameters `parameters = None` and then `if parameters = None: ....` to get rid of the `new` method below?

luxtronik/__init__.py
```diff
@@ -38,31 +39,46 @@ def is_socket_closed(sock: socket.socket) -> bool:
...
+        self.parameters = parameters
+        self.calculations = calculations
+        self.visibilities = visibilities
+
+    @classmethod
+    def new(cls, safe=True):
+        return LuxtronikData(Parameters(safe), Calculations(), Visibilities())
+
+    @classmethod
+    def copy(cls, data):
```
gerw, line 71: No longer needed?

luxtronik/__init__.py
```diff
@@ -233,3 +266,49 @@ def _read_char(self):
...
+        """Read calculations from heat pump."""
+        return self.interface.read_calculations(self.calculations)
+
+    def read_visibilities(self):
+        """Read visibilities from heat pump."""
+        return self.interface.read_visibilities(self.visibilites)
+
+    def write(self, parameters=None):
+        """
+        Write parameter to heat pump.
```
gerw, line 304: I think that this doc string exists now twice?

Guzz-T:
> But this is fixed by moving `Parameters(True)` to the function body in the latest patch, right?
Yes, that has already been adjusted.

Guzz-T:
> Btw: `True` is the default argument of `Parameters()` and can be omitted.
Done

luxtronik/__init__.py
Guzz-T: Done, but still offer a `new(safe)` constructor if someone wants to set the safe parameter.

luxtronik/__init__.py
Guzz-T: This would be necessary if the user wanted to save “old” Luxtronik Data instances.
The he could write `old_data = LuxtronikData.copy(lux.read())`

luxtronik/__init__.py
Guzz-T: Once for `Luxtronik` and once for `LuxtronikSocketInterface`

gerw: I think it would be sufficient to have it only once in the module?

gerw:
The `copy()` method does a shallow copy. We also (instead?) need a deep copy:
````python3
class LuxtronikData:
[...]
    @classmethod
    def deepcopy(cls, data):
        return LuxtronikData(deepcopy(data.parameters), deepcopy(data.calculations), deepcopy(data.visibilities))
````

With this addition, consider this snippet:
````python3
from luxtronik import LuxtronikData
import copy
import timeit

l = LuxtronikData()

l.parameters.get(0).raw = 1

print("l.parameters.get(0).raw =", l.parameters.get(0).raw)

l2 = LuxtronikData.copy(l)
l3 = copy.copy(l)
l4 = copy.deepcopy(l)
l5 = LuxtronikData.deepcopy(l)

print("===")

l.parameters.get(0).raw = 2

print("l.parameters.get(0).raw  =", l.parameters.get(0).raw)
print("l2.parameters.get(0).raw =", l2.parameters.get(0).raw)
print("l3.parameters.get(0).raw =", l3.parameters.get(0).raw)
print("l4.parameters.get(0).raw =", l4.parameters.get(0).raw)
print("l5.parameters.get(0).raw =", l5.parameters.get(0).raw)

print("===")
num = 100
s = """
from luxtronik import LuxtronikData
import copy
l = LuxtronikData()
"""
print(timeit.timeit('l2 = LuxtronikData.copy(l)', setup=s, number=num)/num)
print(timeit.timeit('l3 = copy.copy(l)', setup=s, number=num)/num)
print(timeit.timeit('l4 = copy.deepcopy(l)', setup=s, number=num)/num)
print(timeit.timeit('l5 = LuxtronikData.deepcopy(l)', setup=s, number=num)/num)
````
Output for me:
````
l.parameters.get(0).raw = 1
===
l.parameters.get(0).raw  = 2
l2.parameters.get(0).raw = 2
l3.parameters.get(0).raw = 2
l4.parameters.get(0).raw = 1
l5.parameters.get(0).raw = 1
===
4.6712099992873845e-06
1.3894300013816973e-06
0.00908147751999877
0.009089312020000762
````
Thus:
* The two `copy()` methods do not work as intended.
* The `deepcopy()` are quite slow. On my (rather fast) machine, they need 0.01 seconds.
* The `deepcopy()` from the `copy` module is good enough and we do not need to implement our own `deepcopy()` method.

I think that we can also drop `new()` by using
````python3
class LuxtronikData:
    """
    Collection of parameters, calculations and visiblities.
    Also provide some high level access functions to their data values.
    """

    def __init__(self, parameters=None, calculations=None, visibilities=None, safe=True):
        self.parameters = Parameters(safe) if parameters is None else parameters
        self.calculations = Calculations() if calculations is None else calculations
        self.visibilities = Visibilities() if visibilities is None else visibilities
````

File: luxtronik/__init__.py [hunk 10]
```diff
@@ -159,10 +197,8 @@ def _write(self, parameters):
         parameters.queue = {}
         # Give the heatpump a short time to handle the value changes/calculations:
         time.sleep(WAIT_TIME_AFTER_PARAMETER_WRITE)
```
gerw, line 199: I think this line should be moved to the `_write_and_read` function. We do not need to wait if the user only wants to write.

luxtronik/__init__.py
```diff
@@ -103,26 +116,44 @@ def _with_lock_and_connect(self, func, *args, **kwargs):
...
+        """
+        Read visibilities from heat pump and integrate them to the passed dictionary.
+        This dictionary is returned afterwards, mainly for access to a newly created.
+        """
+        if visibilities is None:
+            visibilities = Visibilities()
+        return self._with_lock_and_connect(self._read_visibilities, visibilities)
 
-    def write(self, parameters):
+    def write(self, parameters, data=None):
```
gerw, line 156: I like the introduction of `_write_and_read()`. I also checked that the last release 0.3.14 did not read the parameters in `write()`. Thus, I propose that `write()` only writes the parameters and another method `write_and_read()` also reads the data after a short wait.

luxtronik/__init__.py
```diff
@@ -233,3 +266,49 @@ def _read_char(self):
...
+
+class Luxtronik(LuxtronikData):
+    """
+    Wrapper around the data and the read/write interface.
+    Mainly to ensure backwards compatibility
+    of the read/write interface to other projects.
+    """
+
+    def __init__(self, host, port=LUXTRONIK_DEFAULT_PORT, safe=True):
+        super().__init__(Parameters(safe), Calculations(), Visibilities())
```
gerw, line 279: With my proposed constructor of `LuxtronikData`, this could be just `super().__init__(safe=safe)`.

luxtronik/__init__.py
Guzz-T: Done

luxtronik/__init__.py
Guzz-T: I would leave it there in case the user write:
```python
lux.write()
lux.read_parameters()
```

Guzz-T:
> The `deepcopy()` from the `copy` module is good enough and we do not need to implement our own `deepcopy()` 
> 
> I think that we can also drop `new()` by using

Done

luxtronik/__init__.py
gerw: But if the user wants to write and read, he can call `write_and_read()`.

luxtronik/__init__.py
Guzz-T: Done

luxtronik/__init__.py
Guzz-T: Done

luxtronik/__init__.py
Guzz-T: Done.

Alternatively, we could implement something like:
```python
def write():
...
    timeout = datetime.now() + timedelta(0,10)

def _lock_and_connect():
    sleep = (timeout - datetime.now()).total_seconds()
    if sleep > 0:
        time.sleep(sleep)
    ...
```

luxtronik/__init__.py
```diff
@@ -103,41 +107,83 @@ def _with_lock_and_connect(self, func, *args, **kwargs):
...
-        prior to reading back in all data from the heat pump.
+        Write all set parameters to the heat pump.
         :param Parameters() parameters  Parameter dictionary to be written
                           to the heatpump before reading all available data
                           from the heat pump.
         """
-        return self._with_lock_and_connect(self._write, parameters)
+        self._with_lock_and_connect(self._write, parameters)
+
+    def write_parameter(self, target, value, safe=True):
```
gerw, line 156: Do we really need this new function?

Guzz-T: In my opinion it is a good addition. Similar to `write_and_read()` where you could also call `write()` and then `read()`, here it is with `parameters.set()` and `write()`.

By the way, the `queue` should actually be independent of the `Parameters` object. But that is not part of this pull request. It would just be a start.

gerw: I am not sure if it is a good idea. In any case, this should not go into this PR.

luxtronik/__init__.py
```diff
@@ -233,3 +272,34 @@ def _read_char(self):
...
+        return self.interface.read_calculations(self.calculations)
+
+    def read_visibilities(self):
+        return self.interface.read_visibilities(self.visibilites)
+
+    def write(self, parameters=None):
+        if parameters is None:
+            self.interface.write(self.parameters)
+        else:
+            self.interface.write(parameters)
```
gerw, line 305: A wrapper for `self.interface.write_and_read` would also be nice.

gerw (commented):
I think that I am (almost) satisfied with the changes in `__init__.py`. (I didn't check the other files yet.)

Thanks for the great work!

luxtronik/__init__.py
Guzz-T: Removed

luxtronik/__init__.py
Guzz-T: Done

gerw:
I have made some changes to the other files. Now, I am fine with the PR

@Guzz-T : Can you please rebase to main and squash all these commits to a single one?

@Bouni @kbabioch : What do you think about the PR?

Guzz-T:
> @Guzz-T : Can you please rebase to main and squash all these commits to a single one?

Done

Bouni:
LGTM I highly like the fact that we unbreak the API 😄

README.md
```diff
@@ -53,27 +53,38 @@ The following example reads in data from the heat pump:
 from luxtronik import Luxtronik
 
 l = Luxtronik('192.168.1.23', 8889)
-calculations, parameters, visibilities = l.read()
+heating_limit = l.parameters.get("ID_Einst_Heizgrenze")
```
gerw, line 56: The ID of the temperature of the heating limit is `ID_Einst_Heizgrenze_Temp`. This is probably meant here. I am not sure, what `ID_Einst_Heizgrenze` is. Maybe a `bool` whether the "Heizgrenze" is enabled?

README.md
```diff
@@ -188,8 +199,15 @@ from luxtronik import Luxtronik, Parameters
 
 l = Luxtronik('192.168.1.23', 8889)
 
+# Queue a parameter change
+l.parameters.set("ID_Einst_Heizgrenze", 14.0)
```
gerw, line 203: This parameter is not marked as `safe` to write, and I think that `14.0` is not a valid value. I suggest to use some parameter, which is actually useful (together with a short comment), what this parameter is about, e.g., `ID_Einst_WK_akt` (temperature offset for heating), `ID_Soll_BWS_akt` (DHW temperature) or `ID_Einst_BWS_Hyst_akt` (DHW hysteresis).

gerw (commented):
IMHO, only two minor things are left. And the commit message "Add a decorator to provide the previous api" might be a copy-and-paste leftover from the last PR?

I am looking forward to merging this.

README.md
Guzz-T: Now use `ID_Soll_BWS_akt`

gerw (approved):

Bouni:
I think everything is resolved and ready. I'll merg it.

Issue #168, opened 2024-02-21
Maintain a list of changed parameter/calculation names


gerw:
Guzz-T noted in #132 that we have changed some names since 0.3.14, in particular, `Unknown_Calculation_239` has been changed to `VBO_Temp_Spread_Soll`. I suspect that this will also happens many times in the future.

Users of our library will find that `Unknown_Calculation_239` will no longer work after an update and in some occasions, it might be quite cumbersome to find out the new name. (For `Unknown_Calculation_239` it is pretty obvious, that this is calculation 239, but in other cases this might not be true).

I think that we should maintain a list of changes of these names, i.e., something like
````python
changed_names = {
    "Unknown_Calculation_239" : "VBO_Temp_Spread_Soll"
}
````
in `constants.py`. Then, `DataVector._lookup` could also resolve old names (and maybe spit out a warning that users can change their code?).

* If we do so, we should also set up a unit test, which assures that all old names can be resolved.
* There seems to be one limit to the approach: We can't resolve the situation where "ID_BarFoo" is mapped to parameter 42 in 0.3.14 but is mapped to 23 in 0.3.15. However, this does not seem to be a big deal, since we can (almost) freely choose the parameter names.
* What about `ID_WEB_SoftStand`? This old calculation will be replaced by a higher level function. I think that an access of this old name should result in a meaningful error message. Maybe we can add a dictionary `obsolete_names` which contains these error messages, e.g.,
````python3
obsolete_names = {
    "ID_Web_SoftStand" : 'This parameter has be replaced by the function "get_firmware_version()" of LuxtronikData / Luxtronik.'
}
````

Guzz-T:
I am aware that this question was a bit "off-topic". However, the library version is usually connected to the firmware version.

It was also more of a question whether it is worth creating a new issue instead of a request to change this here.

PR #171, opened 2024-02-23
Add compatibility feature

gerw:
This adds the suggestion of #168.

The test cases detect if
* a new parameter/calculation/visibility is added, and it also has to be added to the test suite (to make sure that the name is preserved in the future)
* a name has changed.

For obsoleted names, an error is thrown. I am not sure about the test case which checks for the error, because this means that the ID of the obsolete item has to be added to `compatibilities.py`  and to the test suite.


luxtronik/data_vector.py
```diff
@@ -46,6 +51,14 @@ def _lookup(self, target, with_index=False):
             except ValueError:
                 # Get entry by name
                 target_index = None
+
+            if target_index is None:
+                # Resolve compatibility mapping
+                target = LUXTRONIK_COMPATIBILITIES.get(target, target)
```
kbabioch, line 57: Should we log / raise a warning in case you're using an "old" name?

PR #172, opened 2024-02-26
Test the socket interaction by mocking the socket

gerw:
This provides almost 100% coverage of `__init__.py`. I am not sure about the missing lines. Sockets are weird, in particular checking whether a socket is closed / no longer usable.

Also the `struct.error` is not sensible in my opinion, but this should be discussed elsewhere.

*Edit:* And with the tests, I found a typo :)


Bouni:
Nice 😎 

I was tempted to implement tests for `__init__.py` but failed on mocking `socket.socket`, good work @gerw

Bouni (approved):

Issue #174, opened 2024-04-09
Types identified for Parameters 695 and 696


Duke194:
Hey I identified some types of parameters.

695: Unknown("ID_Ba_Hz_MK1_akt"), ---> 695: MixedCircuitMode("ID_Ba_Hz_MK1_akt"),
696: Unknown("ID_Ba_Hz_MK2_akt"), ---> 696: MixedCircuitMode("ID_Ba_Hz_MK2_akt"),

Maybe someone can integrate that, would be very kind :)

gerw:
Included in #175.

PR #175, opened 2024-04-10
Assign MixedCircuitMode datatype to parameters 695 and 696

kbabioch:
This assign the `MixedCircuitMode` datatype to parameters 695 and 696, which have been identified in the context of #174.

It also applies some trivial reformatting to make the ruff linter happy.


gerw (approved):
LGTM

PR #179, opened 2024-06-29
Assign Celsius datatype to visibilities 23 to 40

tfohlmeister:
* visibilities: Assign Celsius datatype to visibilities 23 to 40


gerw:
This does not make sense. All visibilities only take values `0` and `1`. These are definitely not temperatures.

tfohlmeister:
You are right, sorry for that. Please ignore.

PR #181, opened 2024-07-22
Fix reading from socket

gerw:
Sometimes `self._socket.recv` returns fewer bytes than requested (for me, this happens sometimes over VPN). The current `try-check` in `read_*` is not helpful, since it just skews up the readings, e.g., if we only get 2 bytes for parameter 1, the remaining 2 bytes are then feed into parameter 2 and so on...

The current fix assures that the correct amount of bytes is read by introducing a new function `_read_bytes()`.

### DIFF
--- a/luxtronik/__init__.py
+++ b/luxtronik/__init__.py
@@ -205,11 +205,7 @@ class LuxtronikSocketInterface:
         length = self._read_int()
         LOGGER.debug("%s: Length %s", self._host, length)
         for _ in range(0, length):
-            try:
-                data.append(self._read_int())
-            except struct.error as err:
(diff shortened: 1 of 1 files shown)


kbabioch:
What will happen if we request to read more data that is provided by the socket? Can thete be any kind of dead lock in this way?

gerw:
Good point. I think one should check that at least one byte is returned, see https://docs.python.org/3/library/socket.html#socket.socket.recv.

In case nothing is returned by `recv`, the connection is lost. What to do in this case? Raise some error? Which one?

kbabioch (approved):

PR #182, opened 2024-09-20
Fix ruff issues

Bouni:
I noticed that ruff was complaining after I merged a PR that was open for a few months, despite it was marked as all good in the PR view.

Probably som versin mismatch or something like that. 

So I fixed the linter warnings and upgraded the pre-commit-cofig and the lint action.


gerw (approved):
LGTM

PR #183, opened 2024-09-20
Upgrade actions

Bouni:
I noticed this message in the Annotation section of the markdownlint and yamllint actions:

> The following actions use a deprecated Node.js version and will be forced to run on node20: actions/checkout@v3, actions/setup-node@v3. For more info: https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/

So I upgraded all actions versions that have new versions available


gerw (approved):
LGTM

PR #187, opened 2024-10-31
Ensure backwards compatibility

### LINKED ISSUE #168: Maintain a list of changed parameter/calculation names

Guzz-T:
With the changes from these pull requests, the corresponding entries can be found under the old names again. This is possible because you can now assign more than one name to the entries.

The names are passed directly to the constructor. No additional files / additional constants are required. It is therefore immediately clear which names can be used to find the entry.

The current version can now be exchanged with version 0.3.14. Tested with BenPru/luxtronik/2024.10.5-beta on HA 2024.10.4. A new release could then be created (0.4.0?)

Fixes #168.


Bouni (approved):
LGTM

Bouni:
I'll wait for @gerw s review on this before merging it

Guzz-T:
There is one last unresolved issue. I think it comes from the fact that a type `Unknown` was changed to `Timestamp`.

gerw:
I am a little bit biased, because I wrote the previous PR #171. I have had a careful look at both PRs and I must admit that I like my PR more. In my opinion,

* it is cleaner to have the compatibilities in an extra file,
* it is not so nice that `Base._name` has not a 'stable' type (`str` vs `list` of `str`)
* my test cases are more elaborate and make sure that no further incompatibilities arise in the future.

Guzz-T:
> * it is cleaner to have the compatibilities in an extra file,#
Everyone has their own coding style :)


> * it is not so nice that `Base._name` has not a 'stable' type (`str` vs `list` of `str`)
To get around this, you could add the following:
```
if isinstance(name, str):
    self._name = [name]
else:
    self._name = name
```
Then at least `self._name` would be static. 
Or we could always hand over a list. This would then usually only consist of one element.


> * my test cases are more elaborate and make sure that no further incompatibilities arise in the future.
I think we could also use the test code for this.

Guzz-T:
Updated source code to:
- convert `self._names` into a "static" list
- copied test_compatibilities.py from #171 (with minor adjustments: "obsolete" dict removed, visibilities bug added)

gerw:
I still think it is more in the style of this library, if the 
compatibilities reside in their own file. Note that the constants have 
been moved to a separate file some time ago. In the same spirit, the 
compatibilities should be in a separate file and should not "pollute" 
other files.

I do not see the benefit of this new PR over #171

What do the others say in this regard? @Bouni @kbabioch

Guzz-T:
Please note that the current implementation in #171 cannot handle further name changes. As an example, if you change a name from `a` to `b` to `c`. Then you first had to add `a: b`. However, as soon as you switch to `c`, the look-up with `a` no longer works because `b` would then no longer be found either. Maybe you could:
- work with indices
- do the look-up within a loop as long no entry is found


Of course it makes sense to put the constants in an extra file. However, the goal here is to remove or reduce dependencies between the files. 

I see the advantage as:
- reduce complexity
- easy handling
- clearer structure

Bouni:
I just looked at both, this PR and #171 

I like the idea of @Guzz-T to have names as a list and have the first element to be the prefered one and all after that the obsolete names. 

On the other hand, I agree with @gerw that we should have these in a seperate constants file.

How about we merge both approaches?

We have a constants file like this (with a tubple of tuples for each, parameters, calculations and visibilities):

```python
LUXTRONIK_PARAMETER_COMPATIBILITIES = (
    ("ID_Einst_SuSilence", "Unknown_Parameter_1092"), 
    ("ID_Einst_SilenceTimer_0", "Unknown_Parameter_1093"),  
    ("ID_Einst_SilenceTimer_1", "Unknown_Parameter_1094"),
    ("ID_Einst_SilenceTimer_2", "Unknown_Parameter_1095"),
    ...
)
```

Then we change the 3 categories (parameters for example) like so:

```python
...
self._data = {
            0: Unknown(),
            1: Celsius(True),
            2: Celsius(True),
            3: HeatingMode(True),
            4: HotWaterMode(True),
...
```

We then lookup the possible names by the number of the parameter, calculation or visibility.

If the lookup is done via name, we have to iterate over the const until we find the right entry, but that has to be done in either of your approaches anyway if I'm right.

What do you think?

Guzz-T:
With this proposal, it is even more difficult for the user to look up the assignment of the names to the data types.

Should the user work with the constants at all? If not, then I would not implement constants.

gerw:
I would also prefer if `parameters.py` would contain the name, because this is where the user could look up the names. Moreover, it would be nice if this file stays as simple at possible. What about changing `compatibilities.py`to use something like
```python
LUXTRONIK_COMPATIBILITIES = {
    "ID_Einst_SuSilence" : ["Unknown_Parameter_1092", "bar", "foo"],
    ....
}
```
where `ID_Einst_SuSilence` is the current (preferred name) and the list contains the old / obsolete names? Then, there is no need to change `parameters.py`.

Guzz-T:
> I would also prefer if `parameters.py` would contain the name, because this is where the user could look up the names.

For the same reason I would also include the “alternative” names in parameter.py.

only if you insist on outsourcing the “alternative” names. In this case I would like your current suggestion better.

If so, how should we deal with the following:

- identical names in different data vectors (parameter and calculation for example)

- Should we support functions like in this pull-request `Base.get_supported_names()` and `Base.check_name()`

gerw:
> > I would also prefer if `parameters.py` would contain the name, because this is where the user could look up the names.
> 
> For the same reason I would also include the “alternative” names in parameter.py.

But the user does not have to look up the old names?

>     * identical names in different data vectors (parameter and calculation for example)

Is this really an issue? Is there some example where we have a name clash? (I know that 0.3.14 uses `Unknown_Parameters_#` also for visibilities. This seems to be a weak argument, since I do not know any use for the visibilities.)


>     * Should we support functions like in this pull-request `Base.get_supported_names()` and `Base.check_name()`

I am not sure. In my opinion, it would be enough to access the current name of a property, but this should be possible by
```python
parameters.get("outdated_name").name
```

Guzz-T:
> But the user does not have to look up the old names?

I mean, anyone who has worked with this repository is used to looking in the `parameters.py` (or the others) to see what fields and data types there are. The users who already use the old names will wonder why they are no longer there and their code still works (= very magic code).

> Is this really an issue? Is there some example where we have a name clash? (I know that 0.3.14 uses `Unknown_Parameters_#` also for visibilities. This seems to be a weak argument, since I do not know any use for the visibilities.)

That's exactly what I thought of first. See [BenPru's(https://github.com/BenPru/luxtronik/blob/main/custom_components/luxtronik/const.py#L568) usage of `visibilities.Unknown_Parameter_357` which is titled as `V0357_SILENT_MODE_TIME_MENU`. So we would need an entry `"Unknown_Parameter_357": "Silent_mode_time_menu"` for the visibilites and `"Unknown_Parameter_357": "ID_Einst_SuMk225_zeit_0_2"` for the parameters.

Guzz-T:
The longer I think about the topic, I think we also need to support earlier data types. Consider this example:
```python
op_mode = lux.calculations.get("Unknown_Calculation_80")  # outdated name for "ID_WEB_WP_BZ_akt"
op_mode.raw = 1             # op_mode.value returns "1"
if op_mode.value == 1:     # evaluates to "true"
    ...
```

In this example, an object of type `Unknown` is returned. When switching to the latest version, you now get an object of type `OperationMode`. Then:
```python
op_mode = lux.calculations.get("Unknown_Calculation_80") 
op_mode.raw = 1             # op_mode.value returns "hot water"
if op_mode.value == 1:     # evaluates to "false"
    ...
```

This construct would be a possible solution:

```python
self._data = {
     ...
    80: [OperationMode("ID_WEB_WP_BZ_akt"), 
        Unknown("Unknown_Calculation_80")],
    ...
```
or 

```python
self._data = (
     ...
    ([38], OperationMode("ID_WEB_WP_BZ_akt")), 
    ([38], Unknown("Unknown_Calculation_80")),
    ([81..90], IPv4Address("ID_WEB_SoftStand")),
    ...
```

Guzz-T:
Some of the changes in this pull request go beyond just informing the user about the new names. My focus here was on the direct interchangeability of the versions. However, this is not strictly necessary. Do you see it differently?

As a result, I created pull request #191.

Bouni:
Hi everyone, sorry for the long silence. Free time is a rarity in the last months ....


## Regarding name overlaps 

as mentioned by @Guzz-T s question in https://github.com/Bouni/python-luxtronik/pull/187#issuecomment-2461424840

> If so, how should we deal with the following:
> 
>     * identical names in different data vectors (parameter and calculation for example)

I did a quick check and searched all names in all 3 groups for overlaps. There are None so we can rule that out.


## Regarding the lookup issue for the user

As I proposed in #189 I did a first draft in #192 which autogenerates docs for the user to lookup the desired names / ids in a GH pages page: https://bouni.github.io/python-luxtronik/

This gives the user the ability to search through the 3 groups to find the wanted entry.

I could imagine to add a `description` attribute to all datatype classes, that way we could even show that for every entry.
We also could pass a description to all instances, that could also be picked up by the docs generator and show that.

In my opinion its a bad idea to let the user search through the code for the desired name / id.

Concider this a draft with much room for improvement.

## Regarding outdated names

We should print a warning that an outdated name is used and directly show the new name in that message.
That would allow the user to change to the right name very quickly.

I'm also not sure if we should keep that burden of keeping old names at all. Maybe we shoud just raise an Exception if the name is not found.

For me the next release should also bump to the next minor version (0.4.0 I guess) to indicate that this might contain breaking changes.

On the other hand I like the idea of non breaking stuff, but it makes it harder to keep the library slim and easy to understand.

Isn't it enough to have a good changelog which indicates breaking changes!?
To me that feels better than haveing a fairly complicated magic solution under the hood 🤔

gerw:
> We should print a warning that an outdated name is used and directly show the new name in that message. That would allow the user to change to the right name very quickly.
> 
> I'm also not sure if we should keep that burden of keeping old names at all. Maybe we shoud just raise an Exception if the name is not found.


But if we drop the old names, how could the user be informed of the new name?

Guzz-T:
First of all, I like the idea of automatically generated documentation. This is definitely more user-friendly than searching through the code.

> I did a quick check and searched all names in all 3 groups for overlaps. There are None so we can rule that out.

Apart from bug in [visibilities](https://github.com/Bouni/python-luxtronik/blob/0.3.14/luxtronik/visibilities.py#L378), there is currently no overlap. However, this is not a guarantee that this will never happen. 

> We should print a warning that an outdated name is used and directly show the new name in that message.
> That would allow the user to change to the right name very quickly.

Without help/documentation, it is difficult to figure out the new name (Example: Circulation_Pump -> HUP_PWM). Therefore, this should at least be noted in the changelog. An appropriately descriptive error message would, of course, provide even more convenience for the user. In both cases, the change needs to be documented somewhere. Perhaps there is also the possibility of generating a part of the changelog directly from the code and using this piece of code for the error message as well. We should also place value on maintainability here.

> For me the next release should also bump to the next minor version (0.4.0 I guess) to indicate that this might contain breaking changes.

Since there are definitely incompatible changes included, I would even suggest a new major version (1.0.0). There is no reason to be stingy with version numbers.

> The more I think about it, maybe we should disencourage or drop the use of the name all together an only allow the ID.

I don't think it's reliable to trust that the manufacturer won't reorder the values. Therefore, I think that a name-based access to be a good abstraction to catch such changes. Except for a few rare exceptions, the names should be constant and only change for new/unknown values.

Bouni:
> I don't think it's reliable to trust that the manufacturer won't reorder the values. Therefore, I think that a name-based access to be a good abstraction to catch such changes. Except for a few rare exceptions, the names should be constant and only change for new/unknown values.

I highly doubt that they ever change the order of the data coming from the interface. As far as I remember from the days when I reversed the Java App, they don't use the names at all, they purly rely on the order in that the date comes in.
As its only a strem of bytes, the don't have another chance to recognize the data.

If they ever change that order, they will break everything old out there. I'm not sure what relies on this specific interface, but I guess its many things, the AlphaControl App for sure.

I won't say its unthinkable, but at least very unlikely.

I think we should hear the opinions of @gerw @BenPru and @kbabioch on this as well.

Personally I even go as far and think maintainability is slightly higher valued than backwards compatibility.
This library is very very slow in its progress which is due to a certain lack of structure (mostly my fault), so a better maintainanility could make progress faster in my opinion.

kbabioch:
The code itself looks fine / straight-forward to me.

Regarding the preceeding dicussion: I'm fine with merging this change. Whether this aligns with the (current) thinking of @Bouni, @BenPru and @gerw, I don't know.

So I'm happy to merge this, and will give the others some time to look into this and/or veto the change. Otherwise, let's proceed with the merge in a couple of days.

Bouni:
Ok with me, merge it if no one vetos

PR #191, opened 2024-11-20
Throw a KeyError exception if obsolete names are used

### LINKED ISSUE #168: Maintain a list of changed parameter/calculation names

Guzz-T:
Unlike #187, this version no longer supports obsolete names. Instead, it returns an exception, compelling the user to switch to the latest version.

The names are passed directly to the constructor, eliminating the need for additional files or constants. This allows for the identification of names from previous versions without the need to execute the code.

Fixes #168.
Replaces #171 and #187.


Guzz-T:
@Bouni, @gerw How do you like this proposed solution?

kbabioch (approved):

PR #192, opened 2024-12-12
First attempt to auto generate docs

Bouni:
This is an attempt to auto generate a docs page hosted on GH pages

### DIFF
--- a/.github/workflows/scripts/docs/gen-docs.py
+++ b/.github/workflows/scripts/docs/gen-docs.py
@@ -1,0 +1,43 @@
+#!/usr/bin/env python
+import logging
+import re
+from pathlib import Path
+from typing import Literal
+
+from jinja2 import Environment, FileSystemLoader, select_autoescape
+
+logging.basicConfig(level=logging.INFO)
+
+logger = logging.getLogger("doc-gen")
+
+BASEPATH = Path(__file__).resolve().parent
+
+
+def parse_group(group: Literal["calculations", "parameters", "visibilities"]) -> list[dict]:
+    logger.info("parse group '%s'", group)
+    data = []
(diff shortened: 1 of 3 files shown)


Bouni:
I removed the pages branch protection by intent as well as the `on:` branch constraint in order to get a GH pages deüploy that can be live viewed.

https://bouni.github.io/python-luxtronik/

This is a first, very rough draft of what I imagine to be what users use to figure out which id to use in their HA config ....

.github/workflows/scripts/docs/gen-docs.py
```diff
@@ -0,0 +1,43 @@
...
+from jinja2 import Environment, FileSystemLoader, select_autoescape
+
+logging.basicConfig(level=logging.INFO)
+
+logger = logging.getLogger("doc-gen")
+
+BASEPATH = Path(__file__).resolve().parent
+
+
+def parse_group(group: Literal["calculations", "parameters", "visibilities"]) -> list[dict]:
```
Guzz-T, line 16: Isn't it possible to directly import the code and query the data directly, instead of searching the file with regex?

Bouni: Not sure about that to be honest 🤔
Maybe I'll look into that on Monday.
As mentioned, this is mostly a proof of concept, how we collect the data can be changed later on.

Bouni:
@Guzz-T Changed the sccript to use import rather than regex. Obviously much better now, thanks for the input 👍

kbabioch:
What's the current status here? Are you going forward with this @Bouni? There are some minor warnings that might need to be fixed first.

Guzz-T:
I guess this pull-request is outdated and can be closed.

Issue #198, opened 2025-09-15
Feedback on using 'main' (to be used for a 0.3.15 PyPi release) by BenPru/Luxtronik


rhammen:
In this thread I will provide feedback on issues I encountered when using the [current 'main' thread](https://github.com/Bouni/python-luxtronik/tree/bfcf6b29edae5318776408e5c3e937628ef05519) in the HA BenPru/Luxtronik integration, instead of using 0.3.14.

Unless specifically mentioned, it does not mean that these issues need to be fixed before a 0.3.15 version can be released.
Rather, it is intended as feedback on 'Breaking changes' that users may encounter when updating.
It is up to the developers to decide if they want to make changes to (better) deal with reported issue in the code, in documentation, or just accept them.

rhammen:
Several parameters/calculations have been renamed. 
This means that when the old name is used, it results in an error, because the old parameter/calculation is not known anymore.
In most cases this is about calculations/parameters like  "calculations.Unknown_Calculation_XXX"

Some examples, that I countered:

- calculations.Unknown_Calculation_239 --> calculations.VBO_Temp_Spread_Soll
- calculations.Unknown_Calculation_240--> calculations.VBO_Temp_Spread_Ist
- calculations.Unknown_Calculation_242--> calculations.HUP_Temp_Spread_Soll
- calculations.Unknown_Calculation_243--> calculations.Unknown_Calculation_243

There probably are more...

rhammen:
81: Version("ID_WEB_SoftStand"),", which worked and returned the full firmware version (index 81-90) in 0.3.14, does not exist anymore and thus results in an error. 
This calculation has been renamed to "  81: Character("ID_WEB_SoftStand_0"),", and now only returns the first character of the firmware version. 

I did notice the get_firmware_version() function, but users upgrading and encountering this issue would need to look carefully in the code to be aware of it.

PR #204, opened 2025-11-06
Add new parameters/visibilities/calculations names

Guzz-T:
- Add new parameters/visiblities/calculations names as used in [benpru/2025.10.25-beta](https://github.com/BenPru/luxtronik/blob/2025.10.25-beta/custom_components/luxtronik/const.py#L424)
- Add "unknown calculations" names for compatibility reasons

First small step towards #198 

Please feel free to suggest more suitable names. Anything is better than “unknown”.


luxtronik/parameters.py
```diff
@@ -1163,7 +1164,7 @@ def __init__(self, safe=True):
             1116: Unknown("Unknown_Parameter_1116"),
             1117: Unknown("Unknown_Parameter_1117"),
             1118: Unknown("Unknown_Parameter_1118"),
-            1119: Unknown("Unknown_Parameter_1119"),
+            1119: Timestamp(["LAST_DEFROST_TIMESTAMP", "Unknown_Parameter_1119"]),
```
kbabioch: Out of curiosity: Why is this written in all upper case, since most other names are written in a different way?

kbabioch (approved):
Overall looks good to me, just wondering regarding the naming scheme (see previous comment).

luxtronik/parameters.py
Guzz-T: The names in capital letters are used in this way in the HA integration of BenPru. In principle, I would be in favor of a case-insensitive name search (if not already implemented), so that this makes no difference.

PR #208, opened 2025-12-23
Improvements for windows

Guzz-T:
- Remove socket.MSG_DONTWAIT to be able to use the source code under Windows as well.
socket.MSG_DONTWAIT is not supported by Windows

- Call main() in case the module is called via "python -m luxtronik

### DIFF
--- a/luxtronik/__main__.py
+++ b/luxtronik/__main__.py
@@ -25,8 +25,8 @@ def main() -> int:
         description="CLI for Luxtronik controllers",
         usage="""luxtronik <command> [<args>]
         The supported commands are:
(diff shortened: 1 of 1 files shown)


luxtronik/__main__.py
```diff
@@ -47,3 +47,6 @@ def main() -> int:
     sys.argv.pop(1)
     # call the corresponding command
     commands[args.command]()
+
+if __name__ == "__main__":
+    main()
```
kbabioch: Obviously this is a trivial change, still wondering: Is this actually related to "improvements for windows"?

kbabioch (approved):

PR #212, opened 2025-12-23
Return "Unknown_{value}" for not supported selections

### LINKED ISSUE #197: Suggestion: Support unknown codes in SelectionBase
The SelectionBase Class maps a numeric output of the luxtronik controller to a String value.
This works great...., 
... as long as the codes dictionary used for the mapping is up to date. 

Over time, the luxtronik firmware may add additional numeric values to several parameters/calculations that use the SelectionBase class.
Most obvious case is probably the HeatpumpCode class, where new models need to be added. ErrorCodes is another obvious one. 
And are we really sure there are/won't be no other operation modes than currently defined?

Currently, if the numeric value is not present in the codes dictionary, None is returned.
This means that until this library adds the new values to the codes dictionary, users will get a None response.
In the BenPru/Luxtronik issue list I saw several screenshots of users having HeatpumpModel = "None" as a result of this.

I'm not sure returning None the best approach, since it may suggest a problem with the luxtronik controller while it is in fact an outdated codes dictionary in this library. At least it is confusing to the users: why is the Model or ErrorCode None for my heatpump?
My suggestion therefore would be to return (a string of) the numeric value, instead of None, when the numeric value is not found in the dictionary.
Something like f'Unknown_{value}' could also be an option.

I am willing to make a PR for this, but I first like to check how the maintainers or others look at this suggestion.

Guzz-T:
In my opinion, there is no reason to return `None` for unknown values.

Fixes #197

### DIFF
--- a/luxtronik/datatypes.py
+++ b/luxtronik/datatypes.py
@@ -134,6 +134,8 @@ class SelectionBase(Base):
 
     datatype_class = "selection"
 
+    unknown_prefix = "Unknown"
+    unknown_delimiter = "_"
     codes = {}
 
     @classmethod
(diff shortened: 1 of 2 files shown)


kbabioch:
@Bouni Should we proceed with the merge, or do you want to review/proceed here?

Issue #180, opened 2024-07-12
3 New Parameter descriptions found


dasSofa:
I found new Parameter descriptions by tinkering with my 
Alpha Innotec WZSV 62K3M Version V3.89.5 

Channel 252 : T-HG max (Celsius)
Channel 259 : Ausgang 9 (Bool) Output 9 on the Mainboard - used for cooling active Signal for Room thermostat switching between heating and cooling. If you have the Expansion board installed, this Signal comes then from Channel 140 descriped as FP3out.
Channel 260: Freigabe Kühlung (Seconds) enable Cooling 

Channel 267: descriped as "Desired_Room_Temperature" gets me only the constant value 2147483647 .  I dont have an RBE so maybe thats why i dont get a proper temp value there.

Guzz-T:
Note: 2147483647 = 0x7FFF -> value not available

Guzz-T:
ToDo @dasSofa:
- Please check pull request #262 and adjust the definition if necessary
- Test the new data fields in the next release version

Then @Guzz-T :
- Swap the order of the entries for each index
- Add the current field as successor for the new field

PR #262, opened 2026-03-15
Add preliminary calculations fields for 252, 259 and 260

Guzz-T:
This extension does not affect existing queries on the fields used to date. It simply adds new descriptions that can be queried in addition to the existing ones.

Relates to #180

### DIFF
--- a/luxtronik/definitions/calculations.py
+++ b/luxtronik/definitions/calculations.py
@@ -2641,6 +2641,16 @@ CALCULATIONS_DEFINITIONS_LIST: Final = [
     {
         "index": 252,
         "count": 1,
+        "names": ['ID_WEB_Temperatur_THG_max'],
+        "type": Celsius,
+        "writeable": False,
+        "datatype": 'UINT32',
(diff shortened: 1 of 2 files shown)


luxtronik/definitions/calculations.py
```diff
@@ -2638,6 +2638,16 @@
...
     },
+    {
+        "index": 252,
+        "count": 1,
+        "names": ['ID_WEB_Temperatur_THG_max'],
+        "type": Celsius,
+        "writeable": False,
+        "datatype": 'UINT32',
+        "unit": '',
+        "description": 'TODO: Provisional new description based on #180',
```
kbabioch, line 2649: Do we want to merge it like this, or this this something you want to do before merging it?

Guzz-T: I’d like to wait for a reply from @dasSofa to see if the description is correct. Then I’ll merge it and remove the TODO from the description later if everything is in ok and the open points from #180 are done.

PR #264, opened 2026-03-17
Update data fields of the “energy” and “voltage” types

Guzz-T:
- Change the scaling factor of 'ID_Waermemenge' parameters from 0.1 to 0.01
- Change the scaling factor of the voltage data type to 0.01

Relates to #198


Guzz-T:
Can any of you check the values on your heat pump?

kbabioch (approved):
Code itself looks fine to me, but let's give this some time to verify whether it fixes the issue in the real world :-).

Guzz-T:
Are there any findings yet?

rhammen:
@Guzz-T sorry for the late reply. I checked the registers in this PR against the diagnostics dumps collected in the BenPru/luxtronik integration (30 heat pumps from public issue reports plus my own; firmware V1.73–V1.90, V2.88–V2.90, V3.86–V3.92). Result: the 0.01 kWh scale is confirmed for everything here except 1059, which turns out to be firmware-series dependent.

**Confirmed 0.01 kWh — parameter equals the matching calculation (kWh/10) × 100, exactly**

| Parameter | Calculation | Units with data | Example pairs (raw ↔ kWh) |
|---|---|---|---|
| 852 `ID_Waermemenge_Seit` | 154 `ID_WEB_WMZ_Seit` | 9 of 9 | 3449447 ↔ 34494.4, 17613162 ↔ 176131.6, 835590 ↔ 8355.9 |
| 854 `ID_Waermemenge_Hz` | 151 `ID_WEB_WMZ_Heizung` | 26 of 27 | 11960251 ↔ 119602.5, 14545611 ↔ 145456.1, 6449260 ↔ 64492.6 (V1.73) |
| 878 `ID_Waermemenge_BW` | 152 `ID_WEB_WMZ_Brauchwasser` | 24 of 24 | 2615400 ↔ 26154.0, 3004080 ↔ 30040.8, 1337833 ↔ 13378.3 |
| 879 `ID_Waermemenge_SW` | 153 `ID_WEB_WMZ_Schwimmbad` | 5 of 5 | 1565690 ↔ 15656.9, 436410 ↔ 4364.1, 104840 ↔ 1048.4 |

(The one exception for 854 is a ratio of 99.4 from a dump where the two blocks were evidently read a few minutes apart.) This holds on all three series. Note that on V3.88+ units calculation 154 reads 0 while 852 keeps counting — the total is apparently no longer mirrored to the calculation there.

- **1136 / 1137 / 1139 (`HEAT/DHW/COOLING_ENERGY_INPUT`)**: verified against the controller's own display in BenPru/luxtronik#752: 1136 = 543197 vs 5431.9 kWh, 1137 = 222335 vs 2223.3 kWh (LD9, V2.88.3); 1139 = 83272 vs 832.7 kWh (LD5 REV, V2.90.0). Independently, in BenPru/luxtronik#734 a regression of 1136/1137 against 151/152 over 19 snapshots gives a marginal COP of 6.4/3.8 at /100 but 0.64/0.38 at /10.
- **1060 `ID_Waermemenge_Reset`**: same magnitude as 852 and identical to it on 10 of 18 units where it is non-zero (differs in both directions on the other 8, purpose unclear). 0.01 kWh by analogy; no direct verification.

**1059 `ID_Waermemenge_ZWE` — scale depends on the controller series**

- **Series 2 (V2.x): 0.01 kWh** as in this PR. An LD9 on V2.88.3 read 147140 against 1471.4 kWh on the controller's web page (BenPru/luxtronik#752).
- **Series 3 (V3.x): 0.1 kWh.** On five inverter units (MSW4-16, MSW2-6S, MSW2-9S, LW SEC 2; V3.79–V3.92.3) the ZWE energy divided by the ZWE run time equals the rated element power in parameter 1025 only at 0.1 kWh; at 0.01 kWh it would imply a 0.89 kW heating element (BenPru/luxtronik#625).
- Series 1 (V1.x): 1059 reads 0 on every V1 unit in the corpus, so unknown.

So a single scale for 1059 is wrong for one group either way: `Energy` is off by 10× on V2 units, `Energy2` is off by 10× on V3 units (21 of the 30 corpus units are V3). The integration currently handles this with a per-series factor on top of the library, but the library already has what's needed: the `since`/`until` fields on definitions are checked against the firmware version in `Definitions.build()`. I'd propose two entries for 1059:

```python
{"index": 1059, ..., "names": ['ID_Waermemenge_ZWE'], "type": Energy2, "until": '2.99.99', "unit": 'kWh/100', ...},
{"index": 1059, ..., "names": ['ID_Waermemenge_ZWE'], "type": Energy,  "since": '3',       "unit": 'kWh/10',  ...},
```

with the V1 case (unverified, assumed like V2) covered by the first entry. That would make this the first version-gated parameter definition.

**Two more candidates for `Energy2`**, currently `Unknown` upstream and not in this PR — both display-verified in BenPru/luxtronik#752: 1135 = 410381 vs 4103.8 kWh (cooling heat quantity) and 1138 = 113191 vs 1131.9 kWh (pool energy input). 
I'm happy to make PRs for both the 1059 split and the 1135/1138 additions (on top of this one, or separately once it's merged) — or if you'd rather fold them into #264 yourself, that's fine with me too; just say which.

Relates to #198.

PR #265, opened 2026-03-17
Discovered two new error codes

Guzz-T:
image

### DIFF
--- a/luxtronik/datatypes.py
+++ b/luxtronik/datatypes.py
@@ -526,6 +526,8 @@ class Errorcode(SelectionBase):
         805: "sensor desuperheater",
         806: "Modbus SEC",
         807: "Lost modbus connection",
+        810: "Parallel-Operation Version",
+        816: "Multiple SHI data sources",
     }
 
 


luxtronik/datatypes.py
```diff
@@ -526,6 +526,8 @@ class Errorcode(SelectionBase):
         805: "sensor desuperheater",
         806: "Modbus SEC",
         807: "Lost modbus connection",
+        810: "Parallel-Operation Version",
```
kbabioch, line 529: Wondering what this specifically means? Version mis-match in case of parallel operation?

luxtronik/datatypes.py
```diff
@@ -526,6 +526,8 @@ class Errorcode(SelectionBase):
...
+        816: "Multiple SHI data sources",
```
kbabioch, line 530: Is this something that you've triggered during development? Can this still happen with this library when used normally (i.e. not multiple instances, etc.).

Guzz-T: Unfortunately, I’m not exactly sure how I triggered the error. However, this shouldn’t happen during normal use of the library. The classes should implement threads so that concurrent accesses are handled sequentially.

luxtronik/datatypes.py
Guzz-T: I suspect that the firmware versions aren't compatible in a master-slave heatpump setup.

PR #266, opened 2026-03-17
Add new data fields discovered and already used by the HA luxtronik integration

Guzz-T:
The HA luxtronik integration already uses some unknown fields. These changes incorporate the insights gained from this.

Relates to #198


luxtronik/definitions/calculations.py
```diff
@@ -2804,6 +2804,12 @@
         "unit": '°C/10',
         "description": '',
     },
+    {
+        "index": 268,
+        "names": ['Unknown_Calculation_268'],
+        "type": Unknown,
+        "successor": 'AC_Power_Input',
```
kbabioch, line 2811: Out of curiosity: Do we know if this is a measured value, or is this an estimate of the power input? Haven't seen any energy monitoring hardware in my heat pump, so I'm wondering how accurate this could be.

kbabioch (approved):
Changes itself look fine, cannot verify the semantic.

luxtronik/definitions/calculations.py
Guzz-T: I haven't checked that yet. You'd need a reference device that’s accurate enough for that.

Guzz-T:
In fact, only a few names were added. Nothing else should change.

